PR #1, opened 2020-02-05
Basic startup of traefik & drupal labels

jefferya:
First step: basic http port 80 Traefik route to the Drupal container
* fix docker-compose startup problem
* add entry points for 80 and 443
* add traefik labels for drupal

To test:
* clone repo
* ` docker-compose -f docker-compose.mvp1.yml pull`
* ` docker-compose -f docker-compose.mvp1.yml up -d`
* on linux, add `idcp.localdomain` to `/etc/hosts` as an alias for localhost
* ` lynx http://idcp.localdomain:80/`
* if this works, one should see a Drupal 8.8.1 Installation tasks

Where `idcp.localdomain` is the `.env` DOMAIN property

### DIFF
--- a/.env
+++ b/.env
@@ -104,6 +104,17 @@ TOMCAT_LOCALHOST_LOG=WARNING
 TOMCAT_LOCALHOST_MANAGER_LOG=WARNING
 TOMCAT_LOCALHOST_HOST_MANAGER_LOG=WARNING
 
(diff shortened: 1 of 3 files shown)


g7morris:
@jefferya Thanks for this. Looks good, merging. Will work from this template.

PR #7, opened 2020-02-27
Add alpaca connectors and activemq

birkland:
Adds a single Karaf instance containing all connectors, config to run them, and an activemq instance

## To Test
* `docker-compose up -d` (this should pull in all images, if not do a `docker-compose pull` first
* Go to `http://localhost:8161`, which is the ActiveMQ admin console.  Click on "Manage ActiveMQ broker".  Enter in username `admin`, pass `admin`
* Click on the `queues` tab.  You should see all the queues the connectors listen to. 
* Now, pick a service (maybe fits) and remember it.  Shut everything down `docker-compose down -v`
* To to the `.env` file and comment out the env vars for that service.  This will disable it.
* Start up `docker-compose up -d`
* log into ActiveMQ.  Look at the queues, and observe the one that corresponds to your service doesn't exist.  This proves you can shut them off at will

There isn't much more to test, since services aren't connected.  This PR just verifies that camel routes start successfully, connect to the messaging bus and eagerly await messages that never come.

### DIFF
--- a/.env
+++ b/.env
@@ -119,3 +119,47 @@ TRAEFIK_ENTRYPOINTS_WEBSECURE_ADDRESS=:443
 TRAEFIK_API=false
 TRAEFIK_API_DASHBOARD=false
 TRAEFIK_API_INSECURE=false
+
+
+# Activemq address
+ACTIVEMQ_URL=tcp://activemq:61616
+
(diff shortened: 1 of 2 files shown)


g7morris:
@birkland I've successfully tested as advised and am now merging this PR. Thanks! Nice work.

PR #16, opened 2020-04-16
All wired up

dannylamb:
Depends on https://github.com/Islandora-Devops/isle-alpaca/pull/3

Comment in the build sections for and point to ^^ once it's merged (or pull it in and point to a local copy on disk).

## How to Test

From your isle-dc folder:
- `docker-compose down -v`
- `rm -rf codebase && mkdir codebase` to clean the codebase
- If you haven't already done so, `(cd scripts; ./generate_keys.sh)` to generate keys
- `docker-compose up --build -d`
-  `docker exec -it isle_dc_proto_solr bash -c "solr create_core -c islandora"`
- `docker exec -it isle_dc_proto_php bash -c "sh /scripts/islandora/install-islandora.sh"`

You can test by ingesting a repository item, giving it a media and tagging that media as an original file.  This should trigger derivatives.

Admittedly, there's still something super wonky with karaf/camel uploading derivatives, and sometimes you get uploaded files of 0.  I'm still working through that, but you can at least see that the whole system is functioning, albeit a bit impaired.


g7morris:
@dannylamb I've gone ahead and fixed the merge conflict above. I'm going to accept and merge this except for line https://github.com/Islandora-Devops/isle-dc/pull/16/files#diff-4e5e90c6228fd48698d074241c2ba760R223

```bash
    networks:
      internal:
        aliases:
          - idcp.localhost
```

I don't know if this is due to the age of the PR but this hardcoded value for the `traefik` service would introduce issues with the .env in that:

a) I believe this is an older method of calling a domain in docker-compose.yml which isn't optimal

b) would conflict with the current env [PROJECT_BASE_URL=idcp.localhost](https://github.com/Islandora-Devops/isle-dc/blob/development/.env#L8) If someone were to change this value in the .env, this value would conflict. Removing in a followup commit.

PR #67, opened 2020-04-27
Remove blazegraph local bind mount designation

### LINKED ISSUE #66: Blazegraph container fails to start on Windows
Running `make` fails with error:
```
ERROR: for blazegraph  Cannot start service blazegraph: error while mounting volume '/var/lib/docker/volumes/islandora_blazegraph-data/_data': failed to mount local volume: mount /c/Users/Dev/projects/isle-dc/data/blazegraph/data:/var/lib/docker/volumes/islandora_blazegraph-data/_data, flags: 0x1000: no such file or directory
Encountered errors while bringing up the project.
make: *** [Makefile:21: up] Error 1
```

changing the volume type to default (rather than `local`) fixes the issue.  As no other volumes have type `local`, it probably makes sense to bring the blazegraph mount in line line with the others.

birkland:
Brings blazegraph mount in line with the other volumes

Resolves #66

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -343,10 +343,5 @@ volumes:
   activemq-data:
   solr-data:
   blazegraph-data:
-    driver: local
-    driver_opts:
-      type: none
-      o: bind
-      device: ${PWD}/data/blazegraph/data
 
   # isle-dc-postgres-data # Added to prototype but not currently used    


g7morris:
@birkland Thanks for testing this. I'll merge it in for now. Long-term most likely this should be bind-mounted using a local driver. The bigdata.jnl file on large collections can become GBs in size.

PR #70, opened 2020-04-29
Mount codebase as /app for containerized composer

### LINKED ISSUE #69: Makefile cannot build pre-existing codebases
If you checkout an existing Drupal site into `codebase` and do `make`, the build will fail with

```
Composer could not find a composer.json file in /app
To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
make: *** [Makefile:15: drupal_init] Error 1
```

birkland:
This allows the init.sh script to work properly both for creating a new project, and for using an existing project cloned into `codebase`

## To test

To see this problem in `development`, to the following in `isle-dc`
* `git clone https://github.com/Born-Digital-US/drupal-project.git codebase`
* make

You'll see a failure:
```
Composer could not find a composer.json file in /app
To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
make: *** [Makefile:15: drupal_init] Error 1
```

Now, try the same with this PR.  You'll see that `composer.json` is found.  (At present, the `make` process will fail due to a dependency resolution issue, which is irrelevant to this PR)

Resolves #69

### DIFF
--- a/scripts/drupal/init.sh
+++ b/scripts/drupal/init.sh
@@ -44,7 +44,7 @@ function download_drupal() {
 
   # Delete codebase just in case a previous command failed to finish the installation.
   [[ -d ./codebase ]] && rm -rf ./codebase
-  $composer $args $composer_general_flags codebase
+  $composer $args $composer_general_flags .
 
   download_required_packages
 }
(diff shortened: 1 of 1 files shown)


noahwsmith:
I defer to @nikathone on this

scripts/drupal/init.sh
```diff
@@ -149,9 +149,9 @@ if [[ ! $composer ]]; then
   echo >&2
   mkdir -p "$HOME/.composer"
   if $is_darwin; then
-    composer="docker container run -it --rm --user $UID:$GUID -v $HOME/.composer:/tmp -v $PWD:/app composer:1.9.3"
+    composer="docker container run -it --rm --user $UID:$GUID -v $HOME/.composer:/tmp -v $PWD/codebase:/app composer:1.9.3"
   else
-    composer="env MSYS_NO_PATHCONV=1 docker container run -t --rm --user $UID:$GUID -v $HOME/.composer:/tmp -v $PWD:/app composer:1.9.3"
+    composer="env MSYS_NO_PATHCONV=1 docker container run -t --rm --user $UID:$GUID -v $HOME/.composer:/tmp -v $PWD/codebase:/app composer:1.9.3"
   fi
```
nikathone, line 155: I am hesitant to add `codebase` to solve this problem because this will mean composer in this script will only work under `codebase` scope. I think creating a `composer_cmd` function might cover more general use case. See https://gitlab.com/snippets/1971871#LC119 a snippet I just create which move the composer logic into it's own function.

noahwsmith: @nikathone do you want to propose an alternate PR?

birkland: I think another PR would be more appropriate

nikathone: Sure. I will submit one this afternoon.

PR #75, opened 2020-05-06
Swapped out wodby Solr image with official solr. Still getting error …

g7morris:
…in Drupal Status though. Further testing and perhaps config change required.


g7morris:
Unfortunately I'm a bit stuck with this error on the Drupal Status page:

Screenshot from 2020-05-06 09-44-18

After pulling this image settings apart, diffing and reviewing the following Solr settings (schema.xml, solrconfig.xml, stopwords etc) I can report the following:

I can get this image / container to boot, create a new `islandora` core from a bind-mounted islandora_solr_conf configset. However due to permission errors, I've had to incorporate a new command in the Makefile `solr_init` to essentially chmod the `config/solr/islandora_solr_conf` directory with the uid / guid (8983:8983) for the container to launch. 

This also required me to `git config fileMode false` to stop all warnings. 

Additional facts. 

* Running previous wodby image solr configsets of `search_api_solr_8.x-3.21` and a schema name of `drupal-8.3.4-solr-8.x` make the resulting container crash, Unclear why?

* Running Solr 7.1 with a configset of `search_api_solr_8.x-3.2 and a schema / solrconfig name of `drupal-8.3.8-solr-7.x` can get the image working but the same error appears in the Drupal Status page. 

* Diffing `codebase/web/modules/contrib/search_api_solr/solr-conf-templates/7.x` to the results from running islandora-playbook and the current `islandora_solr_conf` results in a lot of, if not all files being the same.

* Oddly, when I attempt to apply the Drupal `config.zip` as normally directed, this config seems to introduce Solr config settings not used in the islandora-playbook or in the previous wodby image. 
In the `solrcore.properties` file, there is this line:
`solr.replication.confFiles=schema.xml,schema_extra_types.xml,schema_extra_fields.xml,elevate.xml,stopwords_en.txt,synonyms_en.txt,protwords_en.txt,accents_en.txt,stopwords_phonetic_und.txt,protwords_phonetic_und.txt,stopwords_phonetic_en.txt,protwords_phonetic_en.txt,stopwords_und.txt,synonyms_und.txt,protwords_und.txt,accents_und.txt
solr.mlt.timeAllowed=2000`
  * Specifically there are about 2-3 txt files with the word `_phonetic` that are only introduced and when attempted to be used i.e. the solrcore.properties file links to them, the container crashes. It is like this config introduces something Solr 8.4.1 chokes on.

Makefile
```diff
@@ -15,7 +15,7 @@ drupal_init:
 	./scripts/drupal/init.sh --codebase $(isle_codebase)
 
 solr_init:
-	./scripts/solr/create-core.sh
+    chown -Rv 8983:8983 config/solr/islandora_solr_conf
```
dannylamb, line 18: You've got a whitespace error that's actually preventing make from understanding your solr_init command.  You actually need a tab here and not whitespace.

dannylamb:
I've managed to hit this, which seems to be preventing it from staying up:
```
Executing /opt/docker-solr/scripts/solr-create -c islandora -d /islandora_solr_conf
Copying solr.xml
cp: not writing through dangling symlink '/var/solr/data/solr.xml'
```
You run into that?

g7morris:
@nikathone I think the question is if this can be fixed, should we move away from using wodby as the Solr image? (I've put you on this for both images to see what it reveals)

PR #77, opened 2020-05-06
Wiring it up

dannylamb:
WIP but everyone may as well track progress on this.

At the moment, this integrates keygen for jwt with the Makefile. and is me toying with the new isle-nginx-php-base and https://github.com/Islandora-Devops/isle-houdini/pull/8


nikathone:
I think we should move `$PWD/config/jwt` under `$PWD/data` if the `jwt` directory will contain keys generated by every user.

g7morris:
@dannylamb I assume you'll merge this in once you determine the correct path?

PR #81, opened 2020-06-04
Bring isle-dc in line with isle-buildkit

noahwsmith:
There are three levels required:

1. Allow local dev
2. Allow spin up of new site from islandora starter
3. Facilitate building remote deploy image

This PR should probably be accepted after we can achieve #1.


noahwsmith:
From @birkland 

> One thing to note is that for isle-dc, we'd eventually want to move away from the paradigm where each service has its own domain name (introduced in isle-buildkit). e.g instead of cantaloupe.localhost, drupal.localhost, fcrepo.localhost, use something akin to isle.localhost/fcrepo, isle.localhost/iiif, etc. Path-based routing instead of host-based. It's fine to use a host-based approach in isle-buildkit, but it's not a pattern folks will use in real life when they deploy an ISLE stack for staging/test/production

noahwsmith:
@dannylamb @g7morris @nigelgbanks @nikathone @bseeger 

Ok, I'm ready to bring this back in. See the docs files under docs/recipes -- the DEMO one works flawlessly as a "boot up Nigel's sandbox" process. The LOCAL one works great for the "bring your own app" variant (because you load your own db which gets around the issue below) but I'm having issues with the "initialize a new app from scratch" which may require changing the "03" script from the Drupal image - it tries to do a site install, and quits because of permissions issues. If it failed silently and stayed up then we could work with it, but s6 just reboots the system infinitely waiting for a database, which we can't install without being able to run drush on the drupal container.

Thoughts?

nigelgbanks (commented):
I spun up the demo locally following the instructions: <https://github.com/noahwsmith/isle-dc/blob/combine-nigel-nia/docs/recipes/DEMO.md>, and got the following error:
```
The website encountered an unexpected error. Please try again later.
```

Seems to be `fedora` is the culprit error appear on <http://islandora.localhost/node/1> after adding a repo item. 

Turns out it was using an older image for sandbox. I needed to pull it as well. I'll put a comment in the relevant spot.

docs/recipes/DEMO.md
```diff
@@ -0,0 +1,41 @@
...
+* `git clone https://github.com/Islandora-Devops/isle-dc.git`
+
+* `cd isle-dc`
+
+* Copy the sample.env file to `.env` 
+  * `cp sample.env .env`
+  * The default `ENVIRONMENT=demo` is correct for this tutorial. You'll switch this to "local" and rebuild if you want to do actual local development (see other related recipes).
+
+* Pull down latest Docker images
+  * `docker-compose pull`
```
nigelgbanks: To ensure their isn't a mismatch between the images we need to pull for all of the `docker-compose.yml` files that will be used.

```
docker-compose -f docker-compose.yml -f docker-compose.demo.yml pull
```

nigelgbanks: Actually seems the make file can do this already:

``
make pull
``

Would be a better fit.

noahwsmith: Yes, that's correct. Not sure if it solves your issue, though? Could you post a gist of your full trace with the error?

nigelgbanks:
In following the local instructions here: <https://github.com/noahwsmith/isle-dc/blob/combine-nigel-nia/docs/recipes/LOCAL.md>

I got the following errors for generating the keys when running ``make install_islandora``:

```bash
Generating RSA Keys
chmod: /opt/keys/claw: No such file or directory
genrsa: Can't open "/opt/keys/claw/private.key" for writing, No such file or directory
Can't open /opt/keys/claw/private.key for reading, No such file or directory
140443857227080:error:02001002:system library:fopen:No such file or directory:crypto/bio/bss_file.c:69:fopen('/opt/keys/claw/private.key','r')
140443857227080:error:2006D080:BIO routines:BIO_new_file:no such file:crypto/bio/bss_file.c:76:
unable to load Private Key
```

The install continued after that point.

Got an error about chmod being called with the incorrect parameters:

```bash
**BusyBox v1.31.1 () multi-call binary.

Usage: chmod [-Rcvf] MODE[,MODE]... FILE...

Each MODE is one or more of the letters ugoa, one of the
symbols +-= and one or more of the letters rwxst

        -R      Recurse
        -c      List changed files
        -v      List all files
        -f      Hide errors
```

Followed by more JWT errors.

```bash
 1/16 [=>--------------------------]   6% [error]  Error getting JWT token for message. Check JWT Configuration. 
 [error]  Error getting JWT token for message. Check JWT Configuration. 
```

As well as some issue with solr which is probably unrelated to the JWT stuff.

```bash
 [error]  Drupal\search_api_solr\SearchApiSolrException while indexing: Solr endpoint http://solr:8983/ not found (404). <html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 404 Not Found</title>
</head>
<body><h2>HTTP ERROR 404</h2>
<p>Problem accessing /solr/ISLANDORA/update. Reason:
<pre>    Not Found</pre></p>
</body>
</html>
 in Drupal\search_api_solr\SolrConnector\SolrConnectorPluginBase->handleHttpException() (line 894 of /var/www/drupal/web/modules/contrib/search_api_solr/src/SolrConnector/SolrConnectorPluginBase.php). 
 [error]  Drupal\search_api_solr\SearchApiSolrException while trying to index items on Default Solr content index: Solr endpoint http://solr:8983/ not found (404). <html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 404 Not Found</title>
</head>
<body><h2>HTTP ERROR 404</h2>
<p>Problem accessing /solr/ISLANDORA/update. Reason:
<pre>    Not Found</pre></p>
</body>
</html>
 in Drupal\search_api_solr\Plugin\search_api\backend\SearchApiSolrBackend->indexItems() (line 998 of /var/www/drupal/web/modules/contrib/search_api_solr/src/Plugin/search_api/backend/SearchApiSolrBackend.php). 
```
The script still returned successful after executing though.

Checked and solr was running:

```fish
docker ps -a | grep solr                                                                                 Wed 17 Jun 2020 15:13:38 UTC
d657ffd2df5c        islandora/solr:latest                 "/init"                  8 minutes ago       Up 8 minutes                                                                            islandora_solr_1
```

I can browse the site at this point but adding content results in the following error message being displayed:

```
Error getting JWT token for message. Check JWT Configuration.
```

The solr issues can be ignored I think as  `make solr_init` after creating the site works and things are getting index. The JWT issue is covered under other comments so **this comment as a whole can be skipped**.

> N.B. I couldn't find the chmod issue in grepping the code and it doesn't seem to be a cause of issues.

docs/recipes/DEMO.md
nigelgbanks: Oh it solved the issue for me, as I had an older sandbox pre Danny's changes to JWT generation.

config/drupal/key.key.islandora_rsa_key.yml
```diff
@@ -12,7 +12,7 @@ key_type_settings:
   algorithm: RS256
 key_provider: file
 key_provider_settings:
-  file_location: /opt/jwt/private.key
+  file_location: /opt/keys/claw/private.key
```
nigelgbanks: Path has changed since https://github.com/Islandora-Devops/isle-buildkit/pull/47

Should now be: ``/opt/keys/jwt/private.key``

scripts/islandora/generate_jwt_keys.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
nigelgbanks, line 1: Behaviour has changed since https://github.com/Islandora-Devops/isle-buildkit/pull/47

It is now an environment/configuration property rather than something that gets generated when the image is run. Though in principal its good to have a script like this to generate the properties for users, it just needs to come in as property.

scripts/islandora/simplified_islandora_install.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
+echo "Generating RSA Keys"
+/opt/scripts/islandora/generate_jwt_keys.sh
```
nigelgbanks, line 3: No longer needed as part of an install, could this instead be moved to the make file and included an env variables?

docker-compose.yml
```diff
@@ -1,287 +1,200 @@
...
-  mariadb:
-    image: mariadb:10.3.22
+  # Single node cluster.
+  # 
+  #  If disabled the system will fall back on Environment Variables.
+  #  To set key/value pairs use the following:
+  #
+  #  ./commands/etcdctl.sh put /drupal/site/default/name "default"
+  #
+  etcd:
```
nigelgbanks, line 52: Should we remove `etcd` from the `docker-compose.yml` file, and instead have users edit an environment file? It will be easier for local development. Also for things like having the makefile generate the public/private keys.

config/drupal/settings.php.local
```diff
@@ -0,0 +1,770 @@
...
+ * caching, JavaScript/CSS compression, re-routing of outgoing emails, and
+ * other things that should not happen on development and testing sites.
+ *
+ * Keep this code block at the end of this file to take full effect.
+ */
+#
+# if (file_exists($app_root . '/' . $site_path . '/settings.local.php')) {
+#   include $app_root . '/' . $site_path . '/settings.local.php';
+# }
+$databases['default']['default'] = array (
```
nigelgbanks, line 758: Seems to be missing the `fedora` configuration for `flysystem`.

I'm not getting any files persisted into Fedora on the local configuration.

noahwsmith:
^ good comments @nigelgbanks . How does this relate to https://github.com/Islandora-Devops/isle-buildkit/pull/52/files - if you're retooling the Drupal install process, maybe we should utilize your new work to solve these issues which largely involve getting scripts to do the work of the sandbox install/setup scripts?

nigelgbanks (changes requested):
I think this can pretty much go in as is, It's really good progress on the developer stories. Issues like `fedora flysystem` can be iterated on afterwards.

The only change I would say should go in before merging is the one line change to `config/drupal/key.key.islandora_rsa_key.yml`.

~Turns out I'm getting a bit confused, the  `docker-compose.local.yml` file is using `islandora/composer` as it's base not `islandora/drupal`. The JWT generation happens in the `islandora/drupal` image and doesn't apply here.~

No I was right originally, got mixed up in that the composer image is two staged and the second stage does use `islandora/drupal`

nigelgbanks:
> ^ good comments @nigelgbanks . How does this relate to https://github.com/Islandora-Devops/isle-buildkit/pull/52/files - if you're retooling the Drupal install process, maybe we should utilize your new work to solve these issues which largely involve getting scripts to do the work of the sandbox install/setup scripts?

That pull only covers a Drupal installation, I haven't cleaned up the `islandora/sandbox` image, yet. If everyone is happy with that general pattern I'll do the same for `islandora/sandbox`.

In the `islandora/sandbox` image we get some duplicated code with `simplified_islandora_install.sh` though the same pattern applies of moving logic out to callable scripts which could be shared across images and pulling in confd settings into the environment to simplify calling those scripts. 

I think maybe in isle-buildkit setting up another image `islandora/islandora` which has the common scripts which could be then shared by downstream images like the sandbox/composer or the custom one in isle-dc, could help simplify the customisation of downstream images.

noahwsmith:
I'd be a big fan of consolidating those initial init/setup scripts into the Drupal container, but then only executing them automatically on the sandbox image. The benefit would be that they'd be available to run manually on the Composer image if anyone wanted to replicate the sandbox's setup without being bound to a sandbox workflow.

And I defer to the sandbox scripts- they're more correct than this project's islandora_install or the _simplified one.

docker-compose.yml
noahwsmith: I was just talking to @g7morris - he agrees and will port over the new sample.env settings from the buildkit repo. We can then kill the etcd service here.

scripts/islandora/simplified_islandora_install.sh
noahwsmith: Yes please. I defer to your other scripts.

scripts/islandora/generate_jwt_keys.sh
noahwsmith: Ok let's kill this if it isn't necessary.

config/drupal/key.key.islandora_rsa_key.yml
noahwsmith: fixed

docs/recipes/DEMO.md
noahwsmith: I fixed this in the docs - thanks

nigelgbanks (approved):

dannylamb (approved):
I managed to run through the Demo instructions.  The `make install_islandora` command needs to be removed from the docs since it runs automatically.  I tried throwing in a `dsm` in the codebase and that isn't coming through, but I think I need to do the "local" build instead of "demo".

Nonetheless, it brought up a site.  I'm :+1: to pulling this in sooner rather than later.  It's a big pull and we still need to iterate to pull out slivers of the setup scripts into the drupal container. So lets do it. :rocket:

noahwsmith:
Terrific- thank you!

PR #83, opened 2020-07-01
Refactored to use changes in simplify/sandbox on buildkit branch.

nigelgbanks:
To be able to test this you need to first **build** the branch in the pull request here: https://github.com/Islandora-Devops/isle-buildkit/pull/55. As those changes are required for this to work.

After that you will need to use those local images rather than the ones from dockerhub. So you'll need to modify the `.env` file to use them.

```bash
make docker-compose.yml
```

Then modify `.env` so that it uses local images.

```bash
REPOSITORY=local
```

At which point you should be good to follow the `README.md` to test the three environments of `demo`, `local`, `custom`.

Once `https://github.com/Islandora-Devops/isle-buildkit/pull/55` is merged users will not have to do the above since the `islandora` images on dockerhub will have the required fixes.


dannylamb (approved):
Let's do this :rocket:

DonRichards:
Is `islandora\demo` a private repo?

PR #86, opened 2020-07-07
Add support for postgresql.

nigelgbanks:
To be able to test this you need to have build: https://github.com/Islandora-Devops/isle-buildkit/pull/65 locally and set `REPOSITORY=local` in your `.env` file.

Changing the following `.env` variables can to `postgresql` will change those services to use postgres:
```
DRUPAL_DATABASE_SERVICE=postgresql
FCREPO_DATABASE_SERVICE=postgresql
GEMINI_DATABASE_SERVICE=postgresql
```
If you building the `local` environment be sure that your `settings.php` file has it's previous database value stripped out of it.

Also this should not be merged until https://github.com/Islandora-Devops/isle-buildkit/pull/65 has been merged since it is a prerequisite.


dannylamb:
I got this going from a clean isle-dc by following the "Create a local codebase from demo" instructions.  Works!

Crazy to be running two dbs but :man_shrugging:

PR #88, opened 2020-07-09
Allow REQUIRED_SERVICES override

birkland:
Primarily for the use case of running ISLE without Fedora, this allows a developer to define the set of services he or she wishes to deploy by way of defining `REQUIRED_SERVICES` in the local `.env` file.


nigelgbanks (approved):
Should do the trick, at a later time maybe separate it out such that fcrepo is included conditionally by a variable if it is not really required.

PR #89, opened 2020-07-09
Tls

dannylamb:
Right now I'm only exposing Drupal, Cantaloupe, and Matomo over https as those are the what would be exposed in a minimal production environment.  Other services are available over regular http, but to change that, you just need to copy/paste/adapt some traefik labels.

To test:
- Pull down https://github.com/Islandora-Devops/isle-buildkit/pull/66 and build the images
- Pull down this PR into your isle-dc repo
- rm docker-compose.yml docker-compose.env.yml .env` to make sure everything is clean.
- `cp sample.env .env`
- Edit `.env` so that `REPOSITORY` is set to `local` so you get the images you just built
- `make`, which will download the default certs and regenerate your files.
- `docker-compose up -d`

Your site should be available at `https://islandora-isle-dc.traefik.me`.

If you are using a remote instance with a domain you own, or an instance with an IP but no domain, follow the instructions in the README to set the proper configuration in your `.env` and `docker-compose.env.yml` files


docker-compose.blazegraph.yml
```diff
@@ -16,10 +16,11 @@ services:
       default:
         # Allow services to use the edge name to reference this service.
         aliases:
-          - blazegraph.${COMPOSE_PROJECT_NAME-isle-dc}.localhost
+          - blazegraph-${COMPOSE_PROJECT_NAME-isle-dc}.${DRUPAL_SITE_HOST-traefik.me}
+          - blazegraph-${COMPOSE_PROJECT_NAME-isle-dc}-${DRUPAL_SITE_HOST-traefik.me}
```
noahwsmith, line 20: duplicated value @dannylamb ?

nigelgbanks:
Code looks great to me, works on localhost for linux, I'll check the other operating systems.

nigelgbanks (approved):
The traefik.me domain doesn't appear to work on windows, though is fine on linux / osx.

If we update the Readme to have users of windows modify their C:\windows\system32\drivers\etc\hosts  it would be fine

I'm happy to merge this as is and do the updating of the read me under another ticket.

nigelgbanks:
Created ticket for updating the read me: https://github.com/Islandora-Devops/isle-dc/issues/90

PR #100, opened 2020-07-22
Single make command to create a dev environment from demo

dannylamb:
Got tired of doing this over and over again, so I scripted it out in the Makefile.  Just run `make dev` and it's the same as if you stepped through the 'create local from demo' steps in the README.


noahwsmith:
Awesomesauce

PR #103, opened 2020-08-08
Some versions of CURL and OpenSSL need fullchain

noahwsmith:
While the simple "cert.pem" file works for some situations, some security settings seem to require the fullchain.pem. There appears to be no downside to using fullchain instead. I left the filename the same so you don't need to edit `tls.yml`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -280,7 +280,7 @@ endif
 download-default-certs:
 	mkdir -p certs
 	if [ ! -f certs/cert.pem ]; then \
-		curl http://traefik.me/cert.pem -o certs/cert.pem; \
+		curl http://traefik.me/fullchain.pem -o certs/cert.pem; \
 	fi
 	if [ ! -f certs/privkey.pem ]; then \
 		curl http://traefik.me/privkey.pem -o certs/privkey.pem; \


dannylamb:
Y'know i wasn't sure about that when I was doing it.  I tried both and opted for the smaller one out of my programmery predisposition.  If this makes it more compatible with http clients i'm all for it.

dannylamb:
I had solved this by updating my local ca-certificates with the fullchain.pem during the throes of development and forgot all about it until just now :sweat_smile: I got rid of it, and re-downloaded the certs using these changes.  Now I can confirm the behaviour.

Before:
```
$ curl -I https://islandora-isle-dc.traefik.me
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

After:
```
$ curl -I https://islandora-isle-dc.traefik.me
HTTP/2 200 
cache-control: must-revalidate, no-cache, private
content-language: en
content-type: text/html; charset=UTF-8
date: Mon, 10 Aug 2020 13:40:07 GMT
expires: Sun, 19 Nov 1978 05:00:00 GMT
server: nginx
x-content-type-options: nosniff
x-drupal-cache: HIT
x-drupal-dynamic-cache: MISS
x-frame-options: SAMEORIGIN
x-generator: Drupal 8 (https://www.drupal.org)
x-powered-by: PHP/7.3.18
x-ua-compatible: IE=edge
```

:+1: Thanks for this @noahwsmith .  cURL not working for folks is just torture.

dannylamb (approved):

DonRichards:
I'm still really confused on how to pull changes in without losing everything

PR #107, opened 2020-08-17
Fixes variables in maria db docker compose file

bseeger:
Fixes typo where some env variables were lower case instead of upper. Resulted in things not working correctly.

### DIFF
--- a/docker-compose.mariadb.yml
+++ b/docker-compose.mariadb.yml
@@ -31,6 +31,6 @@ services:
         # name in settings.php so Drush on the host machine can be used in the
         # codebase folder.
         aliases:
-          - mariadb-${compose_project_name-isle-dc}.${drupal_site_host-traefik.me}
-          - mariadb-${compose_project_name-isle-dc}-${drupal_site_host-traefik.me}
+          - mariadb-${COMPOSE_PROJECT_NAME-isle-dc}.${DRUPAL_SITE_HOST-traefik.me}
+          - mariadb-${COMPOSE_PROJECT_NAME-isle-dc}-${DRUPAL_SITE_HOST-traefik.me}
       gateway:


docker-compose.mariadb.yml
```diff
@@ -31,6 +31,6 @@ services:
         # name in settings.php so Drush on the host machine can be used in the
         # codebase folder.
         aliases:
-          - mariadb-${compose_project_name-isle-dc}.${drupal_site_host-traefik.me}
-          - mariadb-${compose_project_name-isle-dc}-${drupal_site_host-traefik.me}
+          - mariadb-${COMPOSE_PROJECT_NAME-isle-dc}.${DRUPAL_SITE_HOST-traefik.me}
```
noahwsmith, line 34: I'm not great with this syntax, but wouldn't this make more sense?

`- mariadb-${COMPOSE_PROJECT_NAME}-isle-dc.${DRUPAL_SITE_HOST}-traefik.me`

noahwsmith (commented):
I haven't tested this, but I think the variable names are shorter than that. And my bash-fu isn't great, so your code may work by some quirk of syntax, but I would guess mine is easier to read. cc @nigelgbanks for comment

docker-compose.mariadb.yml
bseeger: Yes, I *think* it would.  However it's done this way all over the place and that would be a more global change.  I wonder why it was done that way, though.

nigelgbanks: The bit following the `-` is what will be used if nothing is specified for the value. 

So for example:

```yaml
aliases:
  - mariadb-${COMPOSE_PROJECT_NAME-isle-dc}
```

```bash
COMPOSE_PROJECT_NAME="test" docker-compose config
```

Would output:

```yaml
aliases:
  - mariadb-test
```
And without the variable:
```bash
docker-compose config
```

Would output:

```yaml
aliases:
  - mariadb-isle-dc
```

If we changed the code to use `- mariadb-${COMPOSE_PROJECT_NAME}-isle-dc.${DRUPAL_SITE_HOST}-traefik.me`

Like so: 

```yaml
aliases:
  - mariadb-${COMPOSE_PROJECT_NAME}-isle-dc
```

We'd get:

```bash
COMPOSE_PROJECT_NAME="test" docker-compose config
```

Would output:

```yaml
aliases:
  - mariadb-test-isle-dc
```
And without the variable:

```bash
docker-compose config
```

```yaml
aliases:
  - mariadb--isle-dc
```

nigelgbanks: The change to from `${DRUPAL_SITE_HOST-traefik.me}` to `${DRUPAL_SITE_HOST}-traefik.me` would probably break for non-`traefik.me` use cases, like when people are hosting at their own domain.

nigelgbanks (approved):

Issue #115, opened 2020-10-06
Sed on a mac is different than sed on linux


noahwsmith:
https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux

This means that our Makefile needs those extra ` ''` right after the `-i`.

PR coming right up...

rosiel:
Thanks for the PR - it indeed works on Mac!

Discussion on the Oct 7 Tech call: Gee, we should probably test on Windows too.

@emudojo @birkland ? 

Many thanks! 
(editing to also tag @marksandford  - hi! I hear you're a windows user too?)

PR #116, opened 2020-10-06
Stupid mac tricks: Sed has different syntax on a mac and should be replaced with Perl

noahwsmith:
Addresses #115 - fixes sed for mac. Tested on linux and it still works fine over there, and on mac to ensure `make dev` completes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -292,19 +292,19 @@ dev:
 	$(MAKE) download-default-certs
 	$(MAKE) create-codebase-from-demo
 	if grep -q DRUPAL_DEFAULT_CONFIGDIR docker-compose.env.yml; then \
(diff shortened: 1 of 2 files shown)


noahwsmith:
Ok, I've moved it to Perl from Sed. @dannylamb check me that I'm doing what you intended? I think I got it all working properly.

birkland (dismissed):
I’m not at my machine right now, but I believe Perl would need to be installed explicitly if building on Windows in some circumstances.  So could you add Perl to the list of requirements in the README too?

noahwsmith:
@birkland I make a small change to the Readme to address your request

seth-shaw-unlv (approved):
Works as advertised on my iMac.

seth-shaw-unlv:
Anyone else want to give it a try before we merge this?

rosiel:
I'm running it now... :D so far so good

rosiel (approved):
Yes! this worked!

PR #127, opened 2020-11-12
Mention resource requirements and documentation.

rosiel:
There is no indication in the README that Docker Desktop's default settings need to be changed. Rather than declaring "8GB memory available to Docker" as a requirement (probably isn't), I alluded to it needing to be "available to docker" - hopefully will prevent more people from consistently running into OOM errors and not knowing what to do.

It's clealry documented in the ISLE docs that I wrote, but those aren't acknowledged from the README either, so I tried to place it in context as "a simple walkthrough for a local installation"

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,13 +53,16 @@ supports:
 To quickly get started, we recommend running the [demo](#-demo) environment
 first after you have completed the [Installation](#-installation).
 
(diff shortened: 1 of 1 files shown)


dannylamb (approved):
:+1:

PR #132, opened 2020-12-08
Pushing init logic into Make.  Focusing on 'demo', 'local', and (not q…

dannylamb:
…uite yet but soon) 'remote'


dannylamb:
Ok, so I've got drafts up for everything and it's functionally :+1:. I'm going to work through coding standards etc.. and undraft them and get some testing instructions in here.

dannylamb:
This all works fine if you use secrets.  Unfortunately, I just tested from a fresh clone and bumped into some troubles because by default we're set not to use secrets. It's looking like it's just some more spots to touch up in buildkit with the new defaults. I'll let you know what I find, but for now, if you use secrets everything is :ok_hand:

dannylamb:
I've updated https://github.com/Islandora-Devops/isle-buildkit/pull/95 and have bumped the commit hash to point to it.  This works with and without secrets now.

dannylamb:
This has been updated along with isle-buildkit.  You can try this out by pulling in these changes and running `make demo` to get a sandbox and `make local` to get a barebones install.

nigelgbanks (approved):
Working for me!

PR #137, opened 2021-01-21
Use a commit hash since latest can't be trusted

dannylamb:
This will fix `main`, which is broken right now due to `latest` being unstable.  Just patching the hole in the boat before I make a stable and dev tags


noahwsmith:
I can verify that this solves a lot of problems.

PR #146, opened 2021-03-22
Unbreaking `make local`

dannylamb:
This removes a `composer require drupal/search_api_solr` that snuck into `make local`.  If you build from scratch, you won't neccessarily have `search_api_solr` from the outset.


noahwsmith:
I think @ajstanley reported this. Got a few minutes to test the fix, Alan?

ajstanley:
Works as it's meant to.

PR #153, opened 2021-04-14
New db dump to go along with config export

dannylamb:
Goes with https://github.com/Islandora-Devops/isle-buildkit/pull/140


nigelgbanks:
@dannylamb The commit hash for the other pull hasn't gone along with this though it seems to work with or without that specific commit as a tag so I feel safe in merging it as is, you may want to mint a new release?

PR #156, opened 2021-04-20
Passing through REPOSITORY and TAG variables into `make local`

dannylamb:
Potential fix for an issue twice reported on Slack.  `make local` uses `docker run` to fire up an `nginx` container to use composer and php for stuff. We're neglecting to provide a tag, so it's assuming `latest`, which I have delete from Dockerhub.  These changes should force you to use whatever you've set in the env file.

@seth-shaw-unlv can you confirm this fixes the issue?


seth-shaw-unlv:
🚀

seth-shaw-unlv (approved):
🚀

Issue #165, opened 2021-06-30
Use live and template secrets so git won't complain


Tristaan:
As I have seen this in other projects I recommend this be done here as well.

So what I recommend is that the secrets folder has two subfolders called `live/` and `template/`. The template folder should have all the secrets in it that are required by docker-compose and should be left untouched. Then there should be a make command used to generate secrets inside the live folder (with or without password and certificate generation). The live folder should be excluded with .gitignore. This is quite practical for not commiting local changes to the secrets and no complains from git when pulling new changes when deploying.

I need some feedback if this feature is practical so I can then make it.

noahwsmith:
I like this approach. The only flaw I see is that for some secrets you may want to set a value and have it stick, e.g. https://github.com/Islandora-Devops/isle-dc/blob/development/secrets/DRUPAL_DEFAULT_CONFIGDIR . How would that be accomplished in this structure?

CC @g7morris who I know is also interested in this topic.

Tristaan:
The generator would generate from existing template files and would just
treat some files as exceptions and act accordingly. Nice that you pointed
that out.
So, a Makefile command that generates secrets, this command will have three important variables, one for all the secrets that should be built with a a password generator, one for all the secrets that are certificates (here are a couple of exceptions, note handle). Then the third var will be for secrets which don't require editing and should just be copied in place.

I plan to create this with two for loops and a simple copy command. one for loop for passwords and the other for certificates, which will contain some ifs for the exceptions.
This way must be maintained, because secrets can change. The second way (which requires no maintenance) is to just copy the secrets inplace and leave the editing to the user.

noahwsmith:
Ok- sounds like an excellent approach. I'd say go for it!

Other interested parties may be @dannylamb and @nigelgbanks

dannylamb:
I have no issues with that approach @Tristaan . Sounds better than live editing secrets.

PR #167, opened 2021-07-13
Generate secrets from templates

Tristaan:
Use make generate-secrets to generate secrets for deployment. This is useful against git complaining about changes in secrets.
Also made the secrets more secure so that they cannot be seen by any other users than root.
Also removed some trailing spaces in the makefile.

Reference: #165


Makefile
```diff
@@ -295,14 +295,49 @@ generate-jwt-keys:
...
-endif
 	docker run --rm -ti \
 		--entrypoint php \
 		$(REPOSITORY)/drupal:$(TAG) -r \
 		'echo password_hash(md5("$(MATOMO_USER_PASS)"), PASSWORD_DEFAULT) . "\n";'
 
+.PHONY: generate-secrets
+.SILENT: generate-secrets
+generate-secrets:
+	if [ $$(id -u) -ne 0 ]; then \
```
nigelgbanks: Not super keen on forcing the user to run as `root`, often times they will try some other rule that depends on `generate-secrets` get this message and then apply that for all and it might have unintended consequences. For example `demo` would end up creating `docker-compose.yml` as root, etc.

nigelgbanks:
Side note: I noticed while reviewing I don't think `isle-dc` secrets work with Windows Docker for Desktop? Was reading through the docs and found this: https://docs.docker.com/engine/swarm/secrets/#windows-support.

> Secret files with custom targets are not directly bind-mounted into Windows containers, since Windows does not support non-directory file bind-mounts. Instead, secrets for a container are all mounted in C:\ProgramData\Docker\internal\secrets (an implementation detail which should not be relied upon by applications) within the container. Symbolic links are used to point from there to the desired target of the secret within the container. The default target is C:\ProgramData\Docker\secrets.

Gonna spin off into a separate ticket.

Makefile
```diff
@@ -295,14 +295,49 @@ generate-jwt-keys:
...
-	$(error MATOMO_USER_PASS is not set)
-endif
 	docker run --rm -ti \
 		--entrypoint php \
 		$(REPOSITORY)/drupal:$(TAG) -r \
 		'echo password_hash(md5("$(MATOMO_USER_PASS)"), PASSWORD_DEFAULT) . "\n";'
 
+.PHONY: generate-secrets
+.SILENT: generate-secrets
+generate-secrets:
```
nigelgbanks, line 282: This shares a lot of code with `generate-jwt-keys`. Probably `generate-jwt-keys` should be removed since it's no longer needed.

Makefile
```diff
@@ -295,14 +295,49 @@ generate-jwt-keys:
...
+	if [ $$(id -u) -ne 0 ]; then \
+		echo "generate-secrets must be ran as root."; \
+		exit 1; \
+	fi
+	openssl genrsa -out tmp/private.key 2048 &> /dev/null;
```
nigelgbanks: User may not have `openssl` installed `generate-jwt-keys` used the `drupal` images `openssl` so the user did not have to have it installed.

Makefile
nigelgbanks: Much in the same way as `generate-jwt-keys` uses a docker container to operate as `root` user this could adopt the same approach. For example.

A folder in the repository with a `scripts/generate-jwt-keys.sh`.

```bash
#!/usr/bin/env bash
set -e

readonly PROGNAME=$(basename $0)
readonly PROGDIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
readonly ARGS="$@"

openssl genrsa -out /tmp/private.key 2048 &> /dev/null;
openssl rsa -pubout -in /tmp/private.key -out /tmp/public.key &> /dev/null;
for f in $(find secrets/template/* -printf '%f\n'); do \
if [ $f = "DRUPAL_DEFAULT_CONFIGDIR" ]; then \
	cp secrets/template/$$f secrets/live/$$f; \
	echo "$f copied!"; \
elif [ $$ = "DRUPAL_DEFAULT_SALT" ]; then \
	tr -dc 'A-Za-z0-9-_' </dev/urandom | head -c 74 > secrets/live/$$f; \
	echo "$f generated!"; \
elif [ $f = "JWT_ADMIN_TOKEN" ]; then \
	tr -dc 'A-Za-z0-9' </dev/urandom | head -c 64 > secrets/live/$$f; \
	echo "$f generated!"; \
elif [ $$f = "JWT_PRIVATE_KEY" ]; then \
	cp /tmp/private.key secrets/live/$f; \
	rm /tmp/private.key; \
	echo "$$f generated!"; \
elif [ $f = "JWT_PUBLIC_KEY" ]; then \
	cp /tmp/public.key secrets/live/$f; \
	rm /tmp/public.key; \
	echo "$f generated!"; \
elif [ $f = "MATOMO_USER_PASS" ]; then \
	$(MAKE) generate-matomo-password > secrets/live/$f; \
	echo "$f generated!"; \
else \
	tr -dc 'A-Za-z0-9' </dev/urandom | head -c 48 > secrets/live/$f; \
	echo "$f generated!"; \
fi \
done
sudo chmod 600 secrets/live/*
```

```makfile
.PHONY: generate-secrets
.SILENT: generate-secrets
generate-secrets:
	docker run --rm  \
		-v ./secrets/live:/secrets/live \
		-v ./scripts/scripts/generate-jwt-keys.sh:/generate-jwt-keys.sh \
		--entrypoint bash \
		$(REPOSITORY)/drupal:$(TAG) -s generate-jwt-keys.sh
```

Makefile
```diff
@@ -295,14 +295,49 @@ generate-jwt-keys:
 .PHONY: generate-matomo-password
 .SILENT: generate-matomo-password
 generate-matomo-password:
-ifndef MATOMO_USER_PASS
```
nigelgbanks, line 298: This would make the empty string "" the user password which isn't super secure.

Makefile
```diff
@@ -317,7 +352,7 @@ download-default-certs:
 
 .PHONY: demo
 .SILENT: demo
-demo:
+demo: generate-secrets
```
nigelgbanks, line 304: Generating always will mean the documentation in the read me no longer applies in that user/passwords with defaults will now be random values.

nigelgbanks (changes requested):
Good work and great idea, I've just a few concerns I've raise and I'll send you a pull request with what I think might get around the issues I found.

Tristaan:
Okay, awaiting your PR.

Tristaan:
As a note I want to mention that this idea came from @jjakob

nigelgbanks:
@Tristaan whoops forgot to post the link yesturday https://github.com/Abak-net-d-o-o/isle-dc/pull/1

Tristaan:
Okay, I merged the PR. It is ready now.

nigelgbanks (approved):
Cheers @Tristaan thanks for this!

PR #169, opened 2021-07-20
Create troubleshooting.md

bondjimbond:
@noahwsmith Let me know if this is the wrong location.


noahwsmith:
There's a troubleshooting section in the main Readme.md file with a few items already. Seems like that's a better place? Or move those things in here, and let this be a whole file for troubleshooting tips?

bondjimbond:
I think a separate doc just for troubleshooting (with a reference to it in README.md so it can be easily found) is good

bondjimbond:
OK, some updates to README.md and troubleshooting.md to reflect the above.

noahwsmith:
I support this approach. @dannylamb and @nigelgbanks  - any thoughts?

nigelgbanks (approved):

PR #171, opened 2021-07-23
Expose PHP variables to make them easier to adjust

### LINKED ISSUE #170: Add PHP upload and memory limits to sample.env and docker-compose.drupal.yml
Looking to expose the following PHP variables so they can be quickly and easily changed in ISLE:
```
    PHP_MEMORY_LIMIT
    PHP_POST_MAX_SIZE
    PHP_UPLOAD_MAX_FILESIZE
```

To my understanding, we'll need to add them as variables in `sample.env`, and then reference them in the `environment` section in `docker-compose.drupal.yml` to get them from `.env` into `php.ini`.

I'll make an attempt at this, but also open to someone more competent giving it a go.

bondjimbond:
Resolves #170: Adds PHP memory limit and file upload maximum variables to `sample.env` and `docker-compose.drupal.yml` so that they can be easily adjusted.

### DIFF
--- a/docker-compose.code-server.yml
+++ b/docker-compose.code-server.yml
@@ -14,6 +14,10 @@ secrets:
     file: "./secrets/CODE_SERVER_PASSWORD"
 services:
   code-server:
+    environment:
+      PHP_MEMORY_LIMIT: ${PHP_MEMORY_LIMIT}
+      PHP_POST_MAX_SIZE: ${PHP_POST_MAX_SIZE}
+      PHP_UPLOAD_MAX_FILESIZE: ${PHP_UPLOAD_MAX_FILESIZE}
     image: ${REPOSITORY:-islandora}/code-server:${TAG:-latest}
     labels:
       - traefik.enable=true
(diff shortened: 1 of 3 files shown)


nigelgbanks (changes requested):
Looks solid, existing users will get an error as there is no default value in the *.yml files. Though I don't think it's too much burden for them to check and update their .env when they pull to the latest and compare it against the sample.env.

I think though the same configs need to be added to https://github.com/Islandora-Devops/isle-dc/blob/development/docker-compose.code-server.yml as well. If the user is using the `code-server` it intercepts the traffic that would normally go to `drupal`, and users would likely be confused if they when to the status page / phpinfo in Drupal and didn't see their expected values for memory and what not.

bondjimbond:
@nigelgbanks I believe I addressed that with the latest commit - let me know if it's in the wrong place.

nigelgbanks:
@bondjimbond if you could add it to the `code-server` service as well that would be great!

bondjimbond:
@nigelgbanks Done. Does that address everything?

nigelgbanks (approved):

nigelgbanks:
Fantastic thanks!

Issue #172, opened 2021-07-23
Ability to specify restart policy for containers


misilot:
The ability to specify the restart policy in the .env file would be great!

For example to be able to specify one of the following: no, on-failure, unless-stopped, always

misilot:
Resolved with #173

PR #173, opened 2021-07-23
Add a default restart policy that can be changed

misilot:
Resolves Issue #172

I think I got all the files that would need to be changed.

### DIFF
--- a/docker-compose.activemq.yml
+++ b/docker-compose.activemq.yml
@@ -10,6 +10,7 @@ volumes:
 services:
   activemq:
     image: ${REPOSITORY:-islandora}/activemq:${TAG:-latest}
+    restart: ${RESTART_POLICY:-unless-stopped}
     volumes:
       - activemq-data:/opt/activemq/data
     labels:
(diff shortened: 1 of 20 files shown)


noahwsmith:
This looks beautiful to me. @nigelgbanks / @dannylamb any objections to merging this?

nigelgbanks (approved):

nigelgbanks:
Looks solid, existing users will get an error as there is no default value in the `*.yml` files. Though I don't think it's too much burden for them to check and update their `.env` when they pull to the latest and compare it against the `sample.env`.

misilot:
@nigelgbanks I can update this was a default value. What do you want it to be?

nigelgbanks:
@misilot it should match the same default as is used in `sample.env` for clarity's sake

misilot:
I can do that. I had picked what made sense for me in there but wanted to make sure made sense. I could change the value when redoing this if it didn't make sense for the project.

misilot:
@nigelgbanks a default value has been set

nigelgbanks (approved):

PR #174, opened 2021-07-23
Watchtower vendor change

misilot:
v2tec doesn't have an arm build, so it fails to start on an Apple m1 processor. Switching to the new vendor name appears to work. 

I will be honest though, I do not know how test beyond this simple switch, and that it stays running.

### DIFF
--- a/docker-compose.watchtower.yml
+++ b/docker-compose.watchtower.yml
@@ -4,7 +4,7 @@
 version: "3.7"
 services:
   watchtower:
-    image: v2tec/watchtower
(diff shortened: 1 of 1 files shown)


noahwsmith:
@nigelgbanks you've been working with ARM a bit - how does this look to you?

nigelgbanks:
Looks good to me it was discussed on slack some time ago I pulled the new image and it appeared to work the same as v2tec but with many other features we're not taking advantage of at the moment.

Issue #177, opened 2021-07-30
Document log location


bondjimbond:
I've been having trouble finding the PHP error log and other logs. I don't believe the locations of these logs (both which container they're in and the paths within those containers) are documented anywhere, but they're necessary for troubleshooting. Can we make this information easy to find?

noahwsmith:
The best way to view the logs is as a real-time stream. Run `docker-compose logs -f drupal` to see the Drupal logs

bondjimbond:
Thanks... Is that documented anywhere?

noahwsmith:
I guess not - I suppose it should be mentioned here:

https://github.com/Islandora-Devops/isle-dc/blob/development/docs/troubleshooting.md

PR #179, opened 2021-08-03
Add log info to troubleshooting document

bondjimbond:
Addresses #177


noahwsmith:
I added some more info - thanks for this!

PR #185, opened 2021-10-06
Expose PHP max_execution_time variable to make it easier to adjust

joshdentremont:
Just adding an extra PHP variable based on pull request #171. 

The reason for this is that when I try to export my config files via /admin/config/development/configuration/full/export it times out. Increasing the max_execution_time variable was the only way I could find to fix this.

### DIFF
--- a/docker-compose.code-server.yml
+++ b/docker-compose.code-server.yml
@@ -18,6 +18,7 @@ services:
       PHP_MEMORY_LIMIT: ${PHP_MEMORY_LIMIT}
       PHP_POST_MAX_SIZE: ${PHP_POST_MAX_SIZE}
       PHP_UPLOAD_MAX_FILESIZE: ${PHP_UPLOAD_MAX_FILESIZE}
+      PHP_MAX_EXECUTION_TIME: ${PHP_MAX_EXECUTION_TIME}
     restart: ${RESTART_POLICY:-unless-stopped}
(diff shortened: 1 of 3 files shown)


seth-shaw-unlv:
Could you please provide testing instructions? Thanks.

joshdentremont:
You can see your max execution time by going to https://islandora.traefik.me/admin/reports/status/php and searching for "max_execution_time". By default it is set to 30 seconds, but if you change the variable in your .env and run "make -B docker-compose.yml" you are able to see the change reflected on that page.

The issue with the config export doesn't seem to happen on a stock Islandora install, but once you start adding things and increasing the number of config files it will time out, which requires a longer execution time to be set.

nigelgbanks (approved):
I've tested locally with the instructions from the comments and it works.

red7-soporte:
Hi!

How can I change this php value default_soclet_timeout in isle?
I set PHP_DEFAULT_SOCKET_TIMEOUT=1800 in .env
Then ran:
- make -B docker-compose.yml
- docker-compose up -d

I don't see any change in phpinfo();

joshdentremont:
@red7-soporte Those variables only work if they are also being used in the yml files. For example, that max execution time variable is used here: https://github.com/Islandora-Devops/isle-dc/blob/development/build/docker-compose/docker-compose.drupal.yml#L23

So if you need to change the socket timeout you will likely need to modify the yml file for Drupal.

Issue #184, opened 2021-09-24
"make" not working with docker-compose V2 enabled


pautri:
if the "Use Docker Compose V2 release candidate" experimental feature is enabled (Docker Desktop 4.0.1 Mac Intel), make demo/local fail almost immediately and only a 0 byte docker-compose.yml file is written. I don't recall enabling this myself, it seems to be turned on by default now (see https://github.com/docker/for-mac/issues/5800#issuecomment-896082487)
```
Generating Secrets
Generated or copied ACTIVEMQ_PASSWORD!
Generated or copied ACTIVEMQ_WEB_ADMIN_PASSWORD!
Generated or copied ALPACA_ACTIVEMQ_PASSWORD!
Generated or copied ALPACA_KARAF_ADMIN_PASSWORD!
Generated or copied CODE_SERVER_PASSWORD!
Generated or copied DB_ROOT_PASSWORD!
Generated or copied DRUPAL_DEFAULT_ACCOUNT_PASSWORD!
Generated or copied DRUPAL_DEFAULT_CONFIGDIR!
Generated or copied DRUPAL_DEFAULT_DB_PASSWORD!
Generated or copied DRUPAL_DEFAULT_SALT!
Generated or copied FCREPO_DB_PASSWORD!
Generated or copied GEMINI_DB_PASSWORD!
Generated or copied JWT_ADMIN_TOKEN!
Generated or copied JWT_PRIVATE_KEY!
Generated or copied JWT_PUBLIC_KEY!
Generated or copied MATOMO_DB_PASSWORD!
Generated or copied MATOMO_USER_PASS!
Generated or copied TOMCAT_ADMIN_PASSWORD!
WARN[0000] network gateway: network.external.name is deprecated in favor of network.name
WARN[0000] network gateway: network.external.name is deprecated in favor of network.name
WARN[0000] network gateway: network.external.name is deprecated in favor of network.name
WARN[0000] network gateway: network.external.name is deprecated in favor of network.name
WARN[0000] network gateway: network.external.name is deprecated in favor of network.name
WARN[0000] network gateway: network.external.name is deprecated in favor of network.name
WARN[0000] network gateway: network.external.name is deprecated in favor of network.name
WARN[0000] network gateway: network.external.name is deprecated in favor of network.name
WARN[0000] network gateway: network.external.name is deprecated in favor of network.name
1 error(s) decoding:

* error decoding 'Command': invalid command line string
[...]

PR #186, opened 2021-10-11
Remove deprecated gateway external name

DonRichards:
### Fix warnings

Screen Shot 2021-10-11 at 11 14 22 AM

### DIFF
--- a/docker-compose.activemq.yml
+++ b/docker-compose.activemq.yml
@@ -3,8 +3,7 @@ networks:
   default:
     internal: true
   gateway:
-    external:
-      name: gateway
+    name: gateway
 volumes:
   activemq-data:
 services:
--- a/docker-compose.activemq.yml
+++ b/docker-compose.activemq.yml
@@ -3,7 +3,7 @@ networks:
   default:
     internal: true
   gateway:
-    name: gateway
+    external: true
 volumes:
   activemq-data:
 services:
--- a/docker-compose.blazegraph.yml
+++ b/docker-compose.blazegraph.yml
@@ -3,8 +3,7 @@ networks:
   default:
     internal: true
   gateway:
-    external:
-      name: gateway
+    name: gateway
 volumes:
   blazegraph-data:
 services:
--- a/docker-compose.blazegraph.yml
+++ b/docker-compose.blazegraph.yml
@@ -3,7 +3,7 @@ networks:
   default:
     internal: true
   gateway:
-    name: gateway
+    external: true
 volumes:
   blazegraph-data:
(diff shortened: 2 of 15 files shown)


seth-shaw-unlv:
Can we get testing instructions to reproduce the issue and then demonstrate the fix? Thanks.

DonRichards:
@seth-shaw-unlv Just run `make demo` or some anything to bring it up. The warning should come up. I got this warning on both Linux and MAC. The error is part of the fresh build process. Perhaps run `make clean` first.

DonRichards:
Just noticed another ticket has the warning within the output example. Although his ticket is referencing a composer version issue, it does show the warning messages I posted here. Not completely related but it shows the issue arises when the build process is fired (make local|demo).

https://github.com/Islandora-Devops/isle-dc/issues/184

nigelgbanks:
Looks like the format has [changed](https://docs.docker.com/compose/networking/#use-a-pre-existing-network) so we should, set `external: true` like so:

```yml
gateway:
  external: true
  name: gateway
```

Our current method was deprecated in `version: "3.5"`. In all our `docker-compose` files we have  `version: "3.7"`

It looks like `external: true` must be [specified](https://docs.docker.com/compose/compose-file/compose-file-v3/#external-1) for it to work as it was previously.

nigelgbanks:
Looks like we don't need to specify name either if it matches the same as the network configuration. 

```bash
docker network ls                          Sun 17 Oct 2021 12:54:49 UTC
NETWORK ID     NAME              DRIVER    SCOPE
eb0abd23fac3   bridge            bridge    local
1df64990806d   host              host      local
4943545796cd   isle-buildkit     bridge    local
225f8df1f00d   isle-dc_default   bridge    local
85f08ed21a62   isle-dc_gateway   bridge    local
e540e0060f15   none              null      local
```

So the following would be fine.

```yml
gateway:
  external: true
```

Here we can see that Alpaca is only on the default internal network as is expected and Drupal is on both.

```bash
docker inspect isle-dc_alpaca_1 | jq '.[].NetworkSettings.Networks | keys[]'
"isle-dc_default"
```

```bash
docker inspect isle-dc_drupal_1 | jq '.[].NetworkSettings.Networks | keys[]'
"isle-dc_default"
"isle-dc_gateway
```

nigelgbanks (approved):
I've made the changes I've recommend and tested it locally see the previous comment for the output.

nigelgbanks:
@DonRichards if you're happy with my changes I'll go ahead and merge it.

PR #190, opened 2021-10-27
Make clean doc

DonRichards:
# Simple Pull request

This adds a comment that indicates that `make clean` is an option and how it is different from `docker-compose down -v`

`make clean` was added a while back but no documentation was added for it. This is the doc for it.


nigelgbanks:
Good information to share

PR #194, opened 2021-11-03
Add sudo password warning

DonRichards:
This is to clarify that a password prompt during build/clean isn't asking for a secret but your system's password. This came up during a conversation with someone that had issues building isle-dc.

I had an alternative description if the one was too short. 

> ⚠️ All passwords generated will be automatically added to the docker containers and can be found under `./secrets/live/`. It's likely you will be prompted to enter in a password. This is your local computer's user account password due to an action that requires elevated privileges. It is likely to happen upon building the containers or when running `make clean`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ To get started with a **demo** environment, run:
 ```bash
 make demo
 ```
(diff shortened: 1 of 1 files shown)


DonRichards:
@Islandora-Devops/committers Anyone willing to review this?

seth-shaw-unlv (approved):
Looks good to me. 👍

PR #195, opened 2021-11-15
Add `make help`

DonRichards:
Simple little help command for make. It looks for a comment that starts with double pound `##` with the next line `name:` the name of a function like `up`, `clean`, `down`, `demo`, `local`, `hydrate`, etc. I added a few I thought would be helpful. An example of where to add the `##` and how it sits above the name "config-export:" name line.

```yml
.PHONY: config-export
.SILENT: config-export
## Exports the sites configuration.
config-export:
	docker-compose exec -T drupal drush -l $(SITE) config:export -y
```

## To run/test
```shell
$ make help
```

Screenshot from 2021-11-15 16-11-36

## How does this help
- It creates a dynamic help command
- It helps the process of "self documenting"
- It allows for cherry picking which Make commands to suggest by simply including a `##` doc to imply that a function is potentially intended to be ran by itself.

This does **not** change any other functionality of the makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,7 @@ docker-compose.yml: $(SERVICES:%=docker-compose.%.yml) .env
 	docker-compose $(SERVICES:%=-f docker-compose.%.yml) config > docker-compose.yml
 
 .PHONY: pull
+## Fetches the latest images from the registry.
 pull: docker-compose.yml
 ifeq ($(REPOSITORY), local)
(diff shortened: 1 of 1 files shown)


noahwsmith:
Love it! ❤️

ajstanley (approved):
Works perfectly - this is an excellent addition, and an idea that can be expanded upon

PR #197, opened 2021-11-18
Adds a fix for missing masonry library

DonRichards:
I'm assuming this is a temporary fix and the solution should permanently be ironed out somewhere else but until someone has time this can substitute.

Issues is that the Masonry library is missing upon a fresh build.

### To replicate error message:
Must have something to trigger an update hook and not already have the library installed. This should work from a clean build.
```shell
$ drush updatedb
```

Screenshot from 2021-11-18 11-10-34

### Then run this to fix it:
```shell
# Comment should show 
$ make help
...
  fix-masonry          Fix missing masonry library. <----

$ make fix-masonry
```

If ran a second time it will recognize the directory is there and skip cloning it down.

### Check the fix worked:
Should have no errors prior to entering "y" to import.
```shell
$ drush updatedb 
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -392,5 +392,6 @@ LATEST_VERSION := $(shell curl -s https://api.github.com/repos/desandro/masonry/
 ## Fix missing masonry library.
 fix-masonry:
 	@echo "Latest version of masonry library is ${LATEST_VERSION}"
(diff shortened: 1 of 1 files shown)


DonRichards:
Might try `composer update` to fetch updates that may have a new update hook in it.

DonRichards:
@Islandora-Devops/committers Anyone willing to review this?

seth-shaw-unlv (approved):
I can confirm this works for `make demo` sites.

PR #203, opened 2021-12-17
added options to sample.env for exposing containers via traefik

joshdentremont:
As suggested in [issue 1667](https://github.com/Islandora/documentation/issues/1667) I have added variables to sample.env for each of the services that have traefik.enable labels in their yml files. I have set the defaults to false for everything except Drupal, Matomo, and Canteloupe, since those others should not be exposed in production.

I also added variables for the ports to use when exposing those other containers, with the defaults being the same ports we were using before. For example if you want Solr to run on a port other than 8983 you can change it in your .env and traefik will expose it on the port you supply in the variable.

I also added a variable for the Traefik dashboard which determines api.insecure, api.dashboard, and api.debug all in one.

### DIFF
--- a/docker-compose.activemq.yml
+++ b/docker-compose.activemq.yml
@@ -14,7 +14,7 @@ services:
       - activemq-data:/opt/activemq/data
     labels:
       # Do not expose in production.
-      - traefik.enable=true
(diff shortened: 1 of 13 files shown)


nigelgbanks (approved):

nigelgbanks:
Tested locally and works as described.

PR #207, opened 2022-01-10
Draft - Adds admin theme and menu bar

DonRichards:
Only for "demo" did I add the admin theme improvement. Basically it sets the admin theme to adminimal and adds the adminimal menu so the menus have hover action. This should make the demo experience more user friendly. 

## To test
- Look at any /admin page and try to hover over the menus.
- Pull in the changes and the admin theme will have changed
- Go to to themes and adminimal should be enabled and set to admin default but not the overall default
- Hovering over the menus should now show the contents within the menus


nigelgbanks:
The demo box will soon be coming from https://github.com/Islandora-Devops/islandora-sandbox perhaps that is a more appropriate place for Drupal site changes?

kstapelfeldt:
Needs to be moved after it's done

PR #209, opened 2022-01-10
Increase default timeouts which are insufficient...

jasonhildebrand:
...for generating derivatives.

This is a simple but important PR to improve out-of-the-box usability.  I burned too much time tracking down the cause of this issue!

This PR increases the default timeouts in sample.env.  It also documents the symptoms that end-users might see and how to fix by increasing the limits.


nigelgbanks (approved):

noahwsmith:
I know I'm late on this, but I wanted to comment that this matches our experience for the timeouts (especially Alpaca) that needed to be pushed up. Thanks for pushing this through @jasonhildebrand !

PR #214, opened 2022-01-25
Draft - Add fix-pdfjs

DonRichards:
Running `make local` isn't downloading the PDF.js library. This action was added to the demo container in the buildkit but is missed in the local build. 

This downloads the pdfjs library and sets the permissions in the same manor as the `fix-masonry` command does. 

## To test 
1. Run `make local`
2. Check that '/var/www/drupal/web/libraries/pdf.js' does **not** exists
3. Switch to this pull request
4. Check that '/var/www/drupal/web/libraries/pdf.js' now exists
5. Test PDF.js works.


DonRichards:
In community call it was suggested to move this to a composer install and/or the sandbox. There was still some confusion on the final decision.

PR #219, opened 2022-02-02
Changed defaults for cantaloupe to a working state

nigelgbanks:


DonRichards (approved):
Tested it and it works

PR #226, opened 2022-02-10
Add to for Secrets to be generated before generating secrets

DonRichards:
# Purpose of this:
In a Slack chat it was brought up that users may be running default secrets and not know it and secrets are generated on every build. In fact there was several small improvements that were needed.

Currently the build process will always generate secrets even when **USE_SECRETS** is set to **true** . This adds a simple check then prompt that prevents the code from running the generate script (re)populating the */secrets/live** directory on every make up/down.

For the sake of keeping the description simple I'm going to use the terms `make local`, `make up` or `make demo` interchangeably as a reference to the make build process. It should not matter which of these to use to build ISLE. The tests will not only test the functionality but the ability to bypass/skip a function or not do anything at all. If **USE_SECRETS** is set to **false** or all the secrets exist in the secrets/live/ directory, this will not prompt the user for input.

## 3 scenarios to test for
Test USE_SECRETS set to either true or false for these 3 scenarios. But USE_SECRETS will always result in secrets/live/ being repopulated with new secrets every time a make build process is ran. 

1. Empty secrets/live/ folder
2. secrets/live/ is missing some of the expected secrets
3. Found default values in secrets/live/ secrets

Also because this is part of the build process, you will need to run `make clean` between comparing current behavior vs new and different scenarios.

### To test that this doesn't break existing setups
1. Switch to this PR's branch
2. With Isle either running or not running but previously built with secrets populated run the `make up` command.

You will see a message informing you if you're using default values, otherwise no message.

---
Run `make clean` to reset. Copy sample.env and modify the **.env** file to USE_SECRETS=**false**.

### To test when USE_SECRETS is set to **false**
1. Run `make local`
This should automatically generate custom secrets.

#### To check this worked
Check to see if the directory **/secrets/live** is populated with secrets, most of them are **NOT** set to "**password**"

Run `make clean` to reset. Copy sample.env and modify the **.env** file to USE_SECRETS=**true** to disable new secret generation as the default action.

### Testing that answering **N**o to prompt results in no changes with USE_SECRETS is set to **true**
When the script determines there's an issue, the user is prompted to either generate or copy secrets. This first check verifies nothing happens when **n** or **N** is applied to the prompts.
1. Run `make local`
2. Would you like to generate random secrets? Run a script to create secrets? [y/N] **N**
3. Would you like to copy the default secrets? Run a script to copy secrets? [y/N] **n**
4. Would you like to copy the missing secrets from /home/don/github/isle-dc/secrets/template/? [y/N] **n**
It should exit completely.

#### To check this worked
Check to see if the directory **/secrets/live** is still empty.
 
### Testing auto generate secrets with USE_SECRETS is set to **true**
1. Run `make demo`
2. Would you like to generate random secrets? Run a script to create secrets? [y/N] **y**
This should automatically generate custom secrets just like the last test.
3. Run `make up` again
There should be no message/warnings.

#### To check this worked
Check to see if the directory **/secrets/live** is populated with secrets, most of them are **NOT** set to "**password**"

### Testing copying template secrets with USE_SECRETS is set to **true**
Run `make clean` to reset. Copy sample.env and modify the **.env** file to USE_SECRETS=**true** to disable new secret generation as the default action.

1. Run `make local`
2. Would you like to generate random secrets? Run a script to create secrets? [y/N] **N**
3. Would you like to copy the default secrets? Run a script to copy secrets? [y/N] **y**
This will copy the template/ files to the live/ directory. But it's advised to make changes to these files before proceeding. The messages should be clear (if not please suggest something). This also exits not only the bash script but also the MAKEFILE. This was does intentionally. 
 4. Run 'make up' again
This will check the config and output a message at both the beginning and the end of the build. This message will inform the user which secrets are using default values with information on where to go for further information. 

5. Run `make up` again and the message should come up again last to inform the user they are using default/insecure secrets.

### Testing with missing secrets while USE_SECRETS is set to **true**
1. Delete 1 or many files in **/secrets/live**
2. Run 'make up'
3. Would you like to copy the missing secrets from /home/don/github/isle-dc/secrets/template/? [y/N] **y**
This will copy (not generate) the missing secrets from the template directory.
4. Run `make up` again and message is set back to the warning message when a default value is detected.

Note: This points to where the documentation **should** be but that doesn't exist yet.

# Sorry for the really really long description.


DonRichards:
Added update in the docs to reflect the improvement.

joshdentremont:
@DonRichards I don't think this will work for randomly generated secrets. The other thing USE_SECRETS does is include the file docker-compose.secrets.yml. As I understand it, without this file it won't use secrets at all and must use hard coded values in the containers. This would mean that if USE_SECRETS=false it would never use the randomly generated secrets in secrets/live.

DonRichards:
@joshdentremont Feel free to test this now. I think it covers everything we talked about.

joshdentremont:
@DonRichards sha1sum is not installed on macOS. When I run make local with USE_SECRETS=false I get a bunch of warnings like this:
scripts/check-secrets.sh: line 100: sha1sum: command not found
and it tells me I am using the default secrets

Also, I think we might want to change the name of the USE_SECRETS variable since it no longer does the same thing as it used to and this could confuse people. Maybe we should rename it to USE_CUSTOM_SECRETS or something like that?

DonRichards:
OK, I'll make the change.

DonRichards:
@joshdentremont Give this a try. I switched it to MD5 which is more common and a default for linux mac.

joshdentremont:
Same problem
scripts/check-secrets.sh: line 100: md5sum: command not found

joshdentremont:
Looks like Mac has a program called md5 that does the same thing 
https://stackoverflow.com/questions/8996820/how-to-create-md5-hash-in-bash-in-mac-os-x 
https://security.stackexchange.com/questions/65499/difference-between-os-x-md5-and-gnu-md5sum

DonRichards:
I added a OS check since in linux the command is md5sum and OSX is just md5. I switched the Linux version back to sha1sum and OSX to MD5. Also added a bypass for anyone that wants a CICD script to generate new secrets by passing **yes** into the script directly `bash scripts/check-secrets.sh yes`. It won't impact other functionality. 1 moment and I'll push this. Github is taking a long time.

DonRichards:
OK, @joshdentremont it's ready.

joshdentremont:
It's telling me I'm using default secrets when I'm not. I think this is because of the following 
`> md5 secrets/template/DRUPAL_DEFAULT_ACCOUNT_PASSWORD`
`MD5 (secrets/template/DRUPAL_DEFAULT_ACCOUNT_PASSWORD) = 5f4dcc3b5aa765d61d8327deb882cf99`
`> md5 secrets/template/DRUPAL_DEFAULT_ACCOUNT_PASSWORD | awk '{print $1}'`
`MD5`

I think what you want for MD5 is this (print $4 instead of print $1):
`> md5 secrets/template/DRUPAL_DEFAULT_ACCOUNT_PASSWORD | awk '{print $4}'`
`5f4dcc3b5aa765d61d8327deb882cf99`

DonRichards:
Yeah, made an assumption there. OK. I switched MACs to the failsafe so that I'm not duplicated more code or spending too much time on this. So basically it will compare the cat output for either file. Lets give this one a try.

joshdentremont:
Tested with USE_SECRETS=false and it created the randomly generated secrets but does not use them. I wonder if we should stop if from generating them when it's set to false? This probably doesn't matter but it has confused me in the past.

I also misunderstood how you had set this up so you can ignore my earlier comments about renaming USE_SECRETS. I thought you had made it so USE_SECRETS=false was the way to use randomly generated secrets.

Trying with USE_SECRETS=true and got this issue:
`'Uses Secrets' is set to 'true'.`
`Starting scripts/check-secrets.sh`
`ls: illegal option -- I`
`usage: ls [-@ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1%] [file ...]`
This seems to be another Mac difference, and is set on this line: https://github.com/Islandora-Devops/isle-dc/pull/226/files#diff-019c27b7e1851c31b7dff95fc6645cb2aabe7d3642de7128aca156872fb6721cR61

I think you can just drop the -I because ls will not list hidden files anyway unless you give it the -a option. I removed it and things seem to run as intended

Another issue I see is that I changed only the drupal account password and it's still telling me I am using the default. I don't know if there's a check for individual ones or if it just outputs all the secrets when any of them are using the defaults.

joshdentremont:
Setting `USE_SECRETS=true` then saying no to generating and copying secrets causes the build to quit and no secrets to be generated.

Setting `USE_SECRETS=true` then accepting random secrets seems to work as intended.

Setting `USE_SECRETS=true` then using the defaults seems to work as intended. It exits the first time to give you opportunity to replace the defaults, and if you don't it gives the appropriate warnings

Setting `USE_SECRETS=true` then overwriting the defaults also works as intended and uses the custom secrets without warnings.

I would say this is all working except for the `ls -I` issue and the fact that it seems to be all or nothing with checking whether some secrets are set to the default.

DonRichards:
@joshdentremont I can remove the "i" part of that command. As for changing the admin password, if you're using the `make` to change the password it should update that file. If you're changing the password within Drupal directly it does not output the password to the file and won't be able to track updated passwords like that. 
One of the pros to always having this run upon build is it will always warn the user if they are using defaults and/or if they're missing newly created secrets.

DonRichards:
@joshdentremont Ready to check

joshdentremont:
I am no longer getting that error with ls so that part seems to be working.

I still have the other issue though, I ran make and told it to copy the secrets from the template directory. Then when it exited I changed just the Drupal admin password in the secrets/live directory. Then when I ran make again it set everything up correctly, but still warned me that it was using the default password for Drupal.

`Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/DRUPAL_DEFAULT_ACCOUNT_PASSWORD`

`> cat secrets/live/DRUPAL_DEFAULT_ACCOUNT_PASSWORD`
`abcdefg%`

I can log in to Drupal with the password that I set, so it's working correctly, just giving an incorrect error message.

I didn't change any of the other passwords, just the Drupal one. Is it meant to check each against the template folder individually, or does it just show a warning for all if it finds one default?

DonRichards:
@joshdentremont I can't seem to replicate your issue. When I changed `secrets/live/DRUPAL_DEFAULT_ACCOUNT_PASSWORD` to match your example the output is this.  
```shell
Default Secret -> /home/isle-dc/secrets/live/ACTIVEMQ_PASSWORD
Default Secret -> /home/isle-dc/secrets/live/ACTIVEMQ_WEB_ADMIN_PASSWORD
Default Secret -> /home/isle-dc/secrets/live/ALPACA_ACTIVEMQ_PASSWORD
Default Secret -> /home/isle-dc/secrets/live/ALPACA_KARAF_ADMIN_PASSWORD
Default Secret -> /home/isle-dc/secrets/live/CODE_SERVER_PASSWORD
Default Secret -> /home/isle-dc/secrets/live/DB_ROOT_PASSWORD
Default Secret -> /home/isle-dc/secrets/live/DRUPAL_DEFAULT_DB_PASSWORD
Default Secret -> /home/isle-dc/secrets/live/DRUPAL_DEFAULT_SALT
Default Secret -> /home/isle-dc/secrets/live/FCREPO_DB_PASSWORD
Default Secret -> /home/isle-dc/secrets/live/GEMINI_DB_PASSWORD
Default Secret -> /home/isle-dc/secrets/live/JWT_ADMIN_TOKEN
Default Secret -> /home/isle-dc/secrets/live/JWT_PRIVATE_KEY
Default Secret -> /home/isle-dc/secrets/live/JWT_PUBLIC_KEY
Default Secret -> /home/isle-dc/secrets/live/MATOMO_DB_PASSWORD
Default Secret -> /home/isle-dc/secrets/live/MATOMO_USER_PASS
Default Secret -> /home/isle-dc/secrets/live/TOMCAT_ADMIN_PASSWORD
```
DRUPAL_DEFAULT_ACCOUNT_PASSWORD isn't in the list with a "Default Secret". Would you mind running it again and dropping it into the comments? Is there a chance you mistook "DRUPAL_DEFAULT_DB_PASSWORD" as the same as "DRUPAL_DEFAULT_ACCOUNT_PASSWORD"?

joshdentremont:
```
> make local
'Uses Secrets' is set to 'true'.
Starting scripts/check-secrets.sh

Checking secrets...
  No secrets found in /Users/josh/Developer/library/docker/isle-dc/secrets/live/

There are 2 basic methods to create secrets:
 [1] - Generate new secrets via a script
 [2] - Copy secrets from a /Users/josh/Developer/library/docker/isle-dc/secrets/template directory into /Users/josh/Developer/library/docker/isle-dc/secrets/live/ and then modify them

Would you like to generate random secrets? Run a script to create secrets? [y/N] n

Would you like to copy the default secrets? Run a script to copy secrets? [y/N] y

Copying secrets from /Users/josh/Developer/library/docker/isle-dc/secrets/template/ to /Users/josh/Developer/library/docker/isle-dc/secrets/live/

Suggestion:
    It is much easier to modify these before you start isle than to try to figure out how to push them to the containers.

Exiting build: Please modify the secrets in /Users/josh/Developer/library/docker/isle-dc/secrets/live/ and then run the same make command again.
This is optional, but it is recommended to modify the secrets in /Users/josh/Developer/library/docker/isle-dc/secrets/live/ before running on a production environment.


check-secrets exited 1
make[1]: *** [secrets_warning] Error 1
make: *** [generate-secrets] Error 2

> echo 'qwerty123' > secrets/live/DRUPAL_DEFAULT_ACCOUNT_PASSWORD
> cat secrets/live/DRUPAL_DEFAULT_ACCOUNT_PASSWORD
qwerty123
> make local
'Uses Secrets' is set to 'true'.
Starting scripts/check-secrets.sh
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/ACTIVEMQ_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/ACTIVEMQ_WEB_ADMIN_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/ALPACA_ACTIVEMQ_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/ALPACA_KARAF_ADMIN_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/CODE_SERVER_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/DB_ROOT_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/DRUPAL_DEFAULT_ACCOUNT_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/DRUPAL_DEFAULT_DB_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/DRUPAL_DEFAULT_SALT
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/FCREPO_DB_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/GEMINI_DB_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/JWT_ADMIN_TOKEN
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/JWT_PRIVATE_KEY
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/JWT_PUBLIC_KEY
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/MATOMO_DB_PASSWORD
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/MATOMO_USER_PASS
Default Secret -> /Users/josh/Developer/library/docker/isle-dc/secrets/live/TOMCAT_ADMIN_PASSWORD


	 --- --- WARNING --- ---  --- --- WARNING --- ---


		Using default values for secrets in a production environment is a

					Security Risk

		Default values are identified in /Users/josh/Developer/library/docker/isle-dc/secrets/live/

		If you are using the default values, you can either change the values of
		the file found in /Users/josh/Developer/library/docker/isle-dc/secrets/live/
		Or generate new secrets by running:
			make generate-secrets

		This will generate new secrets in /secrets/live/ but will not update
		the ISLE containers.

		If you are not sure how to push updated secrets to ISLE, please consult
		the documentation.
		https://islandora.github.io/documentation/installation/docker-custom/#secrets


	 --- --- WARNING --- ---  --- --- WARNING --- ---
```

joshdentremont:
I found the issue:
```
		if [[ $hash == "UNKNOWN" ]]; then
			if [[ $(cat secrets/template/ACTIVEMQ_PASSWORD) == $(cat secrets/live/ACTIVEMQ_PASSWORD) ]]; then
				# Ignore the config location directory. This won't pose a security risk.
				if [[ ! "${secret}" = "DRUPAL_DEFAULT_CONFIGDIR" ]]; then
					echo -e "${RED}Default Secret${RESET} ${BLUE}->${RESET} $(pwd)/secrets/live/${secret}"
					FOUND_INSECURE_SECRETS=true
				fi
			fi
		else
			if [[ "$($hash $(pwd)/secrets/template/${secret}| awk '{print $1}')" == "$($hash $(pwd)/secrets/live/${secret}| awk '{print $1}')" ]]; then
				# Ignore the config location directory. This won't pose a security risk.
				if [[ ! "${secret}" = "DRUPAL_DEFAULT_CONFIGDIR" ]]; then
					echo -e "${RED}Default Secret${RESET} ${BLUE}->${RESET} $(pwd)/secrets/live/${secret}"
					FOUND_INSECURE_SECRETS=true
				fi
			fi
		fi
```
It's because I am on Mac and the hash variable is set to UNKNOWN. If we replace `ACTIVEMQ_PASSWORD` with `${secret}` it seems to work fine.

DonRichards:
I'll give that a try and push as soon as I can.

DonRichards:
@joshdentremont OK, ready.

joshdentremont:
@DonRichards Looks good now.

DonRichards:
Ready for testing

misilot:
This worked for me, to not stop on existing secrets being there. ❤️

README.md
```diff
@@ -193,19 +193,24 @@ make up
...
-`make docker-compose.yml`, a large block of `secrets` will be added at the top of
-your `docker-compose.yml` file.
+To enable using secrets prior to running the `make` commands, copy sample.env
+to .env. Set `USE_SECRETS=true` in your .env file. Make a copy of the files in
+/secrets/template/ to /secrets/live/.
+
+To enable using secrets after run `make local` or `make up`, set 
+`USE_SECRETS=true` in your .env file. When you run `make docker-compose.yml`, a
+large block of `secrets` will be added at the top of your `docker-compose.yml`
+file.
```
misilot, line 213: Might want to also mention that they should be sure to set 

to the value of `DRUPAL_DEFAULT_SALT` to $settings['hash_salt'] in codebase/web/sites/default/settings.php and the value of `DRUPAL_DEFAULT_DB_PASSWORD` to the value `$databases['default']['default']['password']` in the same file (it's actually in the file twice??)

Or maybe we should use the file path option instead of parsing it into the file? `$settings['hash_salt'] = file_get_contents('/home/example/salt.txt');`

misilot: Also will a make up update passwords if they are created after a `make local` or `make up`?

misilot (changes requested):

misilot:
I also think it might be worth at somepoint just creating secrets or using existing ones going forward, and not, not having them. ie "USE_SECRETS" goes away except maybe as a legacy option for existing systems? (expert option?)

seth-shaw-unlv:
The last test for the missing secret prompt didn't work when I tested this morning:

```
~/isle-dc$ rm secrets/live/DRUPAL_DEFAULT_DB_PASSWORD
~/isle-dc$ make up
WARNING: Some services (activemq, alpaca, blazegraph, cantaloupe, crayfits, drupal, fcrepo, fits, homarus, houdini, hypercube, mariadb, matomo, milliner, recast, solr, traefik) use the 'deploy' key, which will be ignored. Compose does not support 'deploy' configuration - use `docker stack deploy` to deploy to a swarm.
WARNING: Service "drupal" uses an undefined secret file "/home/sshaw/isle-dc/secrets/live/DRUPAL_DEFAULT_DB_PASSWORD", the following file should be created "/home/sshaw/isle-dc/secrets/live/DRUPAL_DEFAULT_DB_PASSWORD"
```

I should also note that the requests for `make demo` don't result in a working site anyway, but that is probably a separate issue.

README.md
DonRichards: I only see $databases['default']['default']['password'] once in my install.

I just pushed a warning to the user to manually update their settings.php file when it notices a difference between the `secrets/live/DRUPAL_DEFAULT_SALT` and `web/sites/default/settings.php` file. I felt like it might take too long to try to automatically do this for the user and restart their container because this script is running on the local and we'd need to account for Mac command discrepancies.

DonRichards:
@misilot @seth-shaw-unlv This is ready to test again.

seth-shaw-unlv:
In the section "Testing auto generate secrets with USE_SECRETS is set to true" it indicates that after selecting "yes" one should run "make up" again. However, the script simply continues running the "make demo". Is that the expected behavior, @DonRichards?

DonRichards:
No. I don't believe so. Let me double-check now

DonRichards:
@seth-shaw-unlv Sorry, yes, that's correct—this way new people can build right away with as few steps as possible. But if a person is only copying files over, I assume it's for a purpose (to add their passwords to the files).

DonRichards:
Oh, and it should ask the user if they want to exit when they opt for copying the defaults over.

Makefile
```diff
@@ -293,17 +293,23 @@ reindex-triplestore:
 	docker-compose exec -T drupal with-contenv bash -lc 'drush --root /var/www/drupal/web -l $${DRUPAL_DEFAULT_SITE_URL} vbo-exec content emit_node_event --configuration="queue=islandora-indexing-triplestore-index&event=Update"'
 	docker-compose exec -T drupal with-contenv bash -lc 'drush --root /var/www/drupal/web -l $${DRUPAL_DEFAULT_SITE_URL} vbo-exec media emit_media_event --configuration="queue=islandora-indexing-triplestore-index&event=Update"'
 
+# TEST_FILES:=$(wildcard $(CURDIR)/secrets/live/*)
```
seth-shaw-unlv: @DonRichards , do we still need this commented-out line?

seth-shaw-unlv: Other than that, I think we are ready.

README.md
misilot: It looks like it was my older version had some extra things in there

Makefile
DonRichards: It doesn't need to be there. I'll push that really quickly.

seth-shaw-unlv (approved):
👍

PR #228, opened 2022-02-11
Add ability to enable ACME generation of SSL Certs

misilot:
If the user sets `ACME_SERVICE=true`, Traefik will attempt to acquire an SSL Certificate for ${DOMAIN}

Added the following variables with the defaults:

```
USE_ACME=false
ACME_EMAIL-your-email
ACME_SERVER=https://acme-v02.api.letsencrypt.org/directory
TRAEFIK_LOG_LEVEL=ERROR
```

The last 2 variables, I did not add to `sample.env`, do they need to be?

To test this: 

1. Ensure DNS points to custom domain and set it in .env file `DOMAIN=customdomain.com`
2. run `make -B docker-compose.yml`
3. run `make up` or `docker-compose up -d`
4. Go to custom $DOMAIN, and verify you have a valid SSL certificate from Let's Encrypt.

Corresponding documentation update: Islandora/documentation/pull/2051

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,11 @@ ifeq ($(INCLUDE_TRAEFIK_SERVICE), true)
 endif
 
 # The service traefik may be optional if we are sharing one from another project.
+ifeq ($(USE_ACME), true)
+	ACME := acme
+endif
+
+# The service traefik may be optional if we are sharing one from another project.
 ifeq ($(INCLUDE_CODE_SERVER_SERVICE), true)
 	CODE_SERVER_SERVICE := code-server
 endif
@@ -79,7 +84,7 @@ DATABASE_SERVICES := $(sort $(DATABASE_SERVICES))
 # The services to be run (order is important), as services can override one
 # another. Traefik must be last if included as otherwise its network
 # definition for `gateway` will be overriden.
-SERVICES := $(REQUIRED_SERVICES) $(FCREPO_SERVICE) $(WATCHTOWER_SERVICE) $(ETCD_SERVICE) $(DATABASE_SERVICES) $(ENVIRONMENT) $(SECRETS) $(CODE_SERVER_SERVICE) $(TRAEFIK_SERVICE)
+SERVICES := $(REQUIRED_SERVICES) $(FCREPO_SERVICE) $(WATCHTOWER_SERVICE) $(ETCD_SERVICE) $(DATABASE_SERVICES) $(ENVIRONMENT) $(SECRETS) $(CODE_SERVER_SERVICE) $(TRAEFIK_SERVICE) $(ACME)
 
 default: download-default-certs docker-compose.yml pull
 
--- a/.gitignore
+++ b/.gitignore
@@ -91,3 +91,4 @@ Dockerfile
 #===================
 # User provided certs for TLS.
 certs
+acme
--- a/docker-compose.acme.yml
+++ b/docker-compose.acme.yml
@@ -1,0 +1,41 @@
+version: "3.7"
+services:
+  traefik:
+    # Do not set `api.insecure`, `api.dashboard`, `api.debug` to `true` in production.
+    # Also do not expose database 3306/5432, as an entry point.
+    # If the commands below are changed, please copy the changes to `docker-compose.acme.yml`.
+    command:
+      - --api.insecure=${EXPOSE_TRAEFIK_DASHBOARD:-false}
+      - --api.dashboard=${EXPOSE_TRAEFIK_DASHBOARD:-false}
+      - --api.debug=${EXPOSE_TRAEFIK_DASHBOARD:-false}
+      - --entryPoints.http.address=:80
+      - --entryPoints.https.address=:443
+      - --entryPoints.mysql.address=:3306
+      - --entryPoints.postgresql.address=:5432
+      - --entryPoints.fcrepo.address=:8081
+      - --entryPoints.blazegraph.address=:8082
+      - --entryPoints.activemq.address=:8161
+      - --entryPoints.solr.address=:8983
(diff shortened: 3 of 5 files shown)


Makefile
```diff
@@ -74,7 +79,7 @@ DATABASE_SERVICES := $(sort $(DATABASE_SERVICES))
 # The services to be run (order is important), as services can override one
 # another. Traefik must be last if included as otherwise its network
 # definition for `gateway` will be overriden.
-SERVICES := $(REQUIRED_SERVICES) $(FCREPO_SERVICE) $(WATCHTOWER_SERVICE) $(ETCD_SERVICE) $(DATABASE_SERVICES) $(ENVIRONMENT) $(SECRETS) $(CODE_SERVER_SERVICE) $(TRAEFIK_SERVICE)
+SERVICES := $(REQUIRED_SERVICES) $(FCREPO_SERVICE) $(WATCHTOWER_SERVICE) $(ETCD_SERVICE) $(DATABASE_SERVICES) $(ENVIRONMENT) $(SECRETS) $(CODE_SERVER_SERVICE) $(TRAEFIK_SERVICE) $(ACME)
```
wgilling, line 87: @misilot -- I am not an expert in parsing Makefile logic, but this change does not check whether or not `ifeq ($(USE_ACME), true)` as it does above in line 42, and I am wondering (without going through the trouble of testing) if the ACME would need to have an empty / NULL or otherwise default value when not set to USE_ACME. Does this potentially cause the SERVICES to be bad when not USE_ACME? Thanks!!

misilot: @wgilling That's a great question. I will test what happens when set to false (I am pretty sure I did, and it doesn't add in the acme file.)

I used the same logic as the USE_SECRETS variable https://github.com/Islandora-Devops/isle-dc/blob/development/Makefile#L27-L29 

I think what happens is, if it is false or not defined, `$(SECRETS)` just becomes a null value?

misilot: @wgilling 

With `USE_ACME` = false (the default)

```
USE_ACME:  false
ACME:
SERVICES: activemq alpaca blazegraph cantaloupe crayfish crayfits drupal mariadb matomo solr fcrepo6 watchtower drupal.mariadb fcrepo6.mariadb demo secrets traefik
```

With `USE_ACME=true`
```
USE_ACME:  true
ACME:  acme
SERVICES: activemq alpaca blazegraph cantaloupe crayfish crayfits drupal mariadb matomo solr fcrepo6 watchtower drupal.mariadb fcrepo6.mariadb demo traefik acme
```
I believe that addresses your concerns?

misilot:
I just rebased with the latest changes in development

ysuarez:
@misilot I tried to test this on macOS, but I am not sure if I followed the correct steps. 

I first ran
A) `make local` so I would get a `codebase` folder
B) then I updated my /ect/hosts file to resolve a test domain (let's say it was xyz.com) to 127.0.0.1 
C) then I updated the new .ENV' file values you created

```
USE_ACME=true
ACME_EMAIL=newemail@example.com
DOMAIN=xyz.com
```
Then I followed your recommended testing steps...

1. Ensure DNS points to custom domain and set it in .env file `DOMAIN=customdomain.com`
2. run `make -B docker-compose.yml`
3. run `make up` or `docker-compose up -d`
4. Go to custom $DOMAIN, and verify you have a valid SSL certificate from Let's Encrypt.

I was able to load up Islandora with xyz.com (not the actual domain), but I got a cert error. When I opened in the cert in macOS keychain app the domain said it was for traefik.me. Though I had to reset my set up to test another PR, so I want to give this another shot and will save the cert to look at it more closely.

In the meantime how does my workflow look so far? Can this only be tested on an actual server or can I build "make custom" on my macOS machine to try to test this? 

Also, I tried using a ".com" domain that I actually already own, but not sure if that really matters to let's encrypt.

misilot:
@ysuarez the domain needs to resolve to a public IP Address and have appropriate firewall ports open.  Since the default way acme works is it tries and access a file on the webserver (Traefik) over HTTP, and once it verifies that you are in control of the site it will than generate the SSL certificate for you.

ysuarez:
@misilot thanks for explaining how acme works. I was reading up on how "Let's Encrypt" & acme works, but some things were still not clear and at it seemed too easy to abuse. (Though I assumed I was wrong in that impression.) Now it makes a lot more sense that acme being able to connect from outside to verify the domain resolves to a public IP is the key check to proceed with creating the corresponding cert. 

Regretfully at this time I am not able to have the set up needed to test this PR. I was hoping I could help review this PR, AND also learn more about Let's Encrypt. At least I learned a lot thanks to you. Hopefully someone else could test it so this PR gets approved, since it would be a great improvement.

misilot:
Rebased with latest merge

DonRichards (approved):
Passed with some modifications because of a known issue with isle-dc. Commented [here](Best Chainsaw) for modifications. I registered a domain name @ Google Domains and spun up a Linode Ubuntu 20.10 server. The screenshots show the cert results and declared it as valid.

misilot:
@DonRichards 

>Passed with some modifications because of a known issue with isle-dc. Commented [here](Best Chainsaw) for modifications. I registered a domain name @ Google Domains and spun up a Linode Ubuntu 20.10 server. The screenshots show the cert results and declared it as valid.

Thanks for testing! Just curious what modifications did you have to make?

DonRichards:
Look [here](https://github.com/Islandora-Devops/isle-dc/wiki/To-build-isle-dc-on-Linode-from-scratch-(quickly)). The 2 lines following `echo "This is a tmp fix for a known error."`

There was nothing wrong with this PR, it's a known issue with isle-dc currently.

misilot:
Thanks @DonRichards!

PR #234, opened 2022-02-24
Check if term exists before modifying

nigelgbanks:


joshdentremont:
@nigelgbanks This seems to work. I needed to run with this PR for isle-dc and the tag set to the buildkit version that had the drush fixes, but once I used both of those I was able to get the drupal container to run.

nigelgbanks:
@joshdentremont I minted a new release and updated the development branch. https://github.com/Islandora-Devops/isle-dc/pull/236

Issue #235, opened 2022-02-25
Make demo and make local fails if CURDIR has spaces in it


ysuarez:
I believe that if you have spaces in the path for the isle-dc repo on your system, `make demo` and `make local` will fail.

I was using a path similar to this...

`/home/yamil/isle tests/isle-dc`

This of course leads to Make setting the CURDIR to a path with spaces, and then also leads to the makefile generate-secrets target to run commands that fails since it gets parsed as...

```
docker run --rm -t \
	-v /home/yamil/isle tests/isle-dc/secrets:/secrets \
	-v /home/yamil/isle tests/isle-dc/scripts/generate-secrets.sh:/generate-secrets.sh \
	-w / \
	--entrypoint bash \
	islandora/drupal:1.0.0-alpha-11 -c "/generate-secrets.sh && chown -R `id -u`:`id -g` /secrets"
```
(I was able to see how the commands was parsed by running `make generate-secrets -n`)

On macOS Big Sur and Mojave I got this not too obvious error when running `make demo` or `make local`...

```
docker: invalid reference format: repository name must be lowercase.
See 'docker run --help'.
make: *** [Makefile:285: generate-secrets] Error 125
```

Not sure what is best:
1) to wrap relevant uses of $(CURDIR) in the Makefile with double quotes 
2) OR just make it a requirement that the path to isle-dc repo cannot have paths (with a warning in the docs & README). 
3) OR something else entirely

If folks decide to not change the Makefile code I can create a documentation PR to add a note/tip/warning about not having paths with spaces in the README and/or in the official docs. 

For example, in this page that covers [Installing ISLE Docker Compose ](https://islandora.github.io/documentation/installation/docker-prereq/#installing-isle-docker-compose) and/or the section for [Installing a Demo Server](https://islandora.github.io/documentation/installation/docker-demo/#installing-a-demo-server) 

Thoughts? 

Thanks in advance.

ysuarez:
BTW, if folks want to go with the approach of adding quotes to the use of $(CURDIR) I can create a PR for that change.

ysuarez:
At today's tech call I proposed making a PR that will wrap the use of `$(CURDIR)` in double quotes, so that running make commands work even if the full path to isle-dc has spaces in it.

ysuarez:
@DonRichards  (and others) I have been experimenting with Makefile syntax (I had never used Makefiles this complex before), and wondered what you (all) thought of something.

[...]

PR #237, opened 2022-03-04
Changes to allow path to the isle-dc repo to have spaces

ysuarez:
This pull request is an attempt to fix issue #235 to allow the path to the isle-dc repo to have spaces without causing errors when using makefile commands.

Regretfully there are still issues with this pull request in the `drupal-public-files-import` target from its use of the SRC makefile variable.

To test this run `make demo`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ endif
 ## Create Dockerfile from example if it does not exist.
 build:
 	if [ ! -f $(PROJECT_DRUPAL_DOCKERFILE) ]; then \
-		cp "$(CURDIR)"/sample.Dockerfile $(PROJECT_DRUPAL_DOCKERFILE); \
+		cp "$(CURDIR)/sample.Dockerfile" $(PROJECT_DRUPAL_DOCKERFILE); \
 	fi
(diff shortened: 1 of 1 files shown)


ysuarez:
I put the details on the errors I am still getting and what I think are the reasons in issue #235

DonRichards:
The quickest solution is a warning (I just added it). I've tried a few basic methods for escaping the character [space] but have been unsuccessful. It's likely to only take a bit more time but that's additional time than I currently can't invest.

DonRichards:
@ysuarez Do think that will be good enough for now?

ysuarez:
@DonRichards Yes, I think that warning is good enough for now.

If and when when you have time, can you put in here or tell me by Slack what methods you tried that failed so far, so I (or someone else) can use that data to help figure out other ways to try fix the remaining path issues.

ysuarez:
@DonRichards I just tested your changes and the error message was triggered and makes sense. 

```
make demo
Makefile:11: *** Containing path cannot contain space characters: '/path/with/a space/isle-dc'.  Stop.
```
I would support merging this PR , though I am not an ISLE2 developer and "take it with a grain of salt," or at the very least I would suggest that we just add the part that checks if there are spaces in the path.

misilot:
I believe this is limitation with Make per this really old bug report http://savannah.gnu.org/bugs/?712

seth-shaw-unlv (approved):
Works for me. :shipit:

PR #242, opened 2022-03-28
Bugfix: matomo domain env is incorrect

noahwsmith:
Another typo: this variable was changed from early on and it looks like the default Matomo template was never updated.


benglis:
I can confirm this did correctly populate the the config.ini.php in my matomo container with the DOMAIN value specified in my .env file

PR #244, opened 2022-04-05
Fix the TAG/DRUPAL_INSTALL_PROFILE 

DonRichards:
Currently, these are breaking the build. Likely because an install profile PR hasn't been merged yet. This is to set it back to a workable option.

## To test
* Prior to switching to this PR try to build from a clean state. In other words, run `make clean`.
* Run `make up` or `make local`

If it properly builds without issue this would be a successful test.

### DIFF
--- a/sample.env
+++ b/sample.env
@@ -65,7 +65,7 @@ REPOSITORY=islandora
 
 # The version of the isle-buildkit images, non isle-buildkit images have
 # their versions specified explicitly in their respective docker-compose files.
-TAG=aed5adc95784c1dd24589a0f77c3bbb063c92113
+TAG=1.0.0-alpha-15
 
 ###############################################################################
 # Exposed Containers & Ports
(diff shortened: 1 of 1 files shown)


noahwsmith:
#160 was merged. Not sure why this is throwing this error...

DonRichards:
Yeah, I noticed but it introduces a problem. This PR is likely just a temporary fix to just get it to build. I suspect the issue is with the tag. Probably just need a different tag set. Although I prefer the install profile, I can't seem to get "islandora_install_profile_demo" to build properly from a clean state without making these changes. From what I'm seeing in chat and PR reviews this has become a blocker.

misilot:
@noahwsmith I think it might be because of this PR https://github.com/Islandora-Devops/isle-buildkit/pull/170 and not #160

seth-shaw-unlv (approved):
This resolves the build issues I was experiencing. We'll run with it until the referenced build-kit PR gets in.

PR #247, opened 2022-04-07
Load env changes with single make command

DonRichards:
Keeping with the idea of helping the user from needing to know/use docker/docker-compose commands and simplifying the process.

This adds `make env` to fetch the changes made to the .env file post-build. 

## To test
Run `make clean` to start with a clean environment then run `make env`

It should fail with a message "No .env file found." and exit with doing nothing to the system.

- Now run `make local` to build the system. This will automatically create a .env file. 
- Navigate to http://islandora.traefik.me:8983/solr/#/ and this should give you a 404. Watch out! Your browser might try to redirect you to the HTTPS equivalent. This will not work. Switch it back to HTTP manually.
- Modify the .env file to expose solr
- Run `make help` and the output should include " env _ Pull in changes to the .env file.`
- Run `make env` and navigate to http://islandora.traefik.me:8983/solr/#/

Also added some text formatting to the help definitions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -454,9 +454,8 @@ confirm:
 	@echo -n "Are you sure you want to continue and drop your data? [y/N] " && read ans && [ $${ans:-N} = y ]
 
 RESET=$(shell tput sgr0)
-RED=$(shell tput setaf 1)
-GREEN=$(shell tput setaf 2)
-YELLOW=$(shell tput setaf 3)
+RED=$(shell tput setaf 9)
+BLUE=$(shell tput setaf 6)
 TARGET_MAX_CHAR_NUM=20
 
 .PHONY: help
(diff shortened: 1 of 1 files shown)


ysuarez:
(Here are my initial tests results. I will do more testing later and add another comment.)

A) I tested this PR and did not get `make env` to show the "No .env file found" message because lines 3-7 of the Makefile always create an empty `.env` file if one is missing every time the Makefile runs...

https://github.com/Islandora-Devops/isle-dc/blob/189b0be3ad884631c3dd76938ccfbf049c7b5945/Makefile#L3-L7

If I comment out lines 3-7 then I do get the expected error message.

B) the text formatting to the help definitions looks great!

ysuarez:
Here is the second part of my testing, and the PR worked as expected.

- I ran `make local` to build the system, and got the expected Solr page 404. 
- Modified the .env file to expose solr
- Ran `make env` and navigated to http://islandora.traefik.me:8983/solr/#/ and saw the Solr admin page. (Though there were no cores.)

image

There is a small potential bug when running `make clean` then make `make env`, but besides that this PR works fine.

DonRichards:
@ysuarez It should never create an empty .env file. It's supposed to copy the sample.env file. And yes, that no cores thing is a `make local` issue/feature.

ysuarez (approved):
This PR works as expected. I think it should be committed.

BTW Don, I wrote something wrong in an earlier comment. I didn't mean to say we would get an empty `.env` file, I meant that with one of the suggested testing instructions of running `make clean` then `make env`, I would not end up with a missing `.env` to trigger the warning `.env` error message in the PR that the file was missing. This is because currently whenever any make command is run as you mentioned, lines 3-7 of the Makefile will automatically create an `.env` that is a copy of the `sample.env`.  I think it is a good defensive practice by this PR to be thorough in first checking if the `.env` exists, before doing its work.

PR #252, opened 2022-04-21
Updates Traefik to Latest Release

misilot:
Provides some new features for ACME and a bunch of other fixes and features

I tested this by updating my docker-compose.traefik.yml file and doing a `make up` and the services all appeared to be there and running.

I am looking to increase some of the ACME settings and need at least v2.6.0

Thanks!

# Related Tickets
https://github.com/Islandora-Devops/isle-dc/pull/253
https://github.com/Islandora-Devops/isle-dc/pull/252
https://github.com/Islandora/documentation/pull/2096

### DIFF
--- a/docker-compose.traefik.yml
+++ b/docker-compose.traefik.yml
@@ -15,7 +15,7 @@ networks:
 services:
   traefik:
     restart: ${RESTART_POLICY:-unless-stopped}
-    image: traefik:2.2.1
+    image: traefik:2.7
     container_name: traefik
     # Do not set `api.insecure`, `api.dashboard`, `api.debug` to `true` in production.
     # Also do not expose database 3306/5432, as an entry point.


ysuarez:
@misilot I have been researching how to use zerossl, an it turns out they disallow using the free EC2 domains AWS gives you...

https://help.zerossl.com/hc/en-us/articles/360060863094-Forbidden-domain-names

I need to see if I can reuse any unused domains I may own, and have and point it to an AWS instance. Then I think I can create a free zerossl 90 day SSL cert. Though I will need more time to test this. I wonder if I can use a DigitalOcean instance with zerossl, not sure if DigitalOcean offers a "free domain" for their instances. @DonRichards I know you have used DigitalOcean for ISLE2 tests before, do you think that I should try using DigitalOcean for testing this PR? I have not used them yet, but I can try them out.

misilot:
Ah, sorry didn't realize that! Sorry about that @ysuarez.

If you have access to a domain you could just create  a random subdomain like `islandora.domainname.com` and point it towards your EC2 instance. It doesn't need to be the apex domain.

ysuarez:
I started looking at adding a third-level domain on a personal domain that I already own, but my hosting company did not let me. I will ping you later on Slack to brainstorm on ideas on how to test this. For example, I was wondering if I need to use an AWS Elastic IP address for a test EC2 instance. Since I think at some point I have to update a name server to point the third-level domain to the EC2 instance, but I may not want to keep the EC2 instance running long enough or the DNS changes to propagate, hence I may target an Elastic IP. Though I could be completely mistaken.

misilot:
I can provide a subdomain to test if that would be helpful. We can talk on slack.

DonRichards:
@misilot Are we needing to test subdomains before we merge this? The scope of this looks like it's only upgrading traefik from 2.2.1 to 2.6.6, is that correct?

misilot:
@DonRichards, if islandora.traefik.me works, I think it should be good.

I would also update to 2.7.0, since that was released recently. 

But yes, the scope of this was only upgraded Traefik.

misilot:
@DonRichards what are your thoughts on just targeting v2.7 instead of a specific point release?

misilot:
@DonRichards I just figured out your question. I believe @ysuarez's question may belong to #253

seth-shaw-unlv (approved):
I also had to update `docker-compose.yml` as well for the change to take, but I was able to view the site and generate derivatives, so everything looks good to me. 👍

misilot:
Sorry about that @seth-shaw-unlv. Missed the `make -B docker-compose.yml` step. 

Thanks for merging!

PR #259, opened 2022-05-17
Reordering the Fedora Port in docker-compose file

misilot:
Since Fedora does need to be exposed publicly, moving it above the warning to not expose it.


noahwsmith:
Thanks Tom!

PR #263, opened 2022-06-01
Added template prototype for endusers using TLS 1.2 & 1.3, multiple domains or Let's Encrypt.

g7morris:
Added template prototype for endusers using TLS 1.2 & 1.3, multiple domains or Let's Encrypt.


tls.yml
```diff
@@ -1,4 +1,19 @@
 tls:
+# For TLS 1.2+ and multiple domains and ssls
```
misilot: @DonRichards I would think we would want to use these as the default options for isle-dc, to encourage best practices.

tls.yml
```diff
@@ -7,3 +22,15 @@ tls:
   certificates:
     - certFile: /etc/ssl/traefik/cert.pem
       keyFile: /etc/ssl/traefik/privkey.pem
+# Uncomment and edit accordingly for remote certs - commercial prod & staging example.
```
misilot: I think these are good, so people know this is all they need to change, but probably should put a note that they need to be put into the `certs` directory.

tls.yml
```diff
@@ -7,3 +22,15 @@ tls:
   certificates:
     - certFile: /etc/ssl/traefik/cert.pem
       keyFile: /etc/ssl/traefik/privkey.pem
+# Uncomment and edit accordingly for remote certs - commercial prod & staging example. 
+    #- certFile: /etc/ssl/traefik/commercial_wildcard.crt
+    #  keyFile: /etc/ssl/traefik/commercial_wildcard.key
+# Uncomment and update docker-compose traefik labels accordingly with the resolver name e.g. myresolver
```
misilot: So, we already have https://github.com/Islandora-Devops/isle-dc/blob/development/docker-compose.acme.yml that can be turned on and off via the `.env` file. So not sure if we want to include this here since it would add in the configuration in two different places. 


This is in reference to the `certificatesResolvers:` section (incase line numbers change)

tls.yml
misilot: Though, I wonder if this could be set via an environmental variable in `.env`? The only reason I am thinking this, is this is a file in SCM, so not something that one would expect to have to edit for localized configuration.

g7morris: Good call @misilot I updated the comment to direct folks to put files there.

tls.yml
g7morris: I can remove if that is the call.

tls.yml
DonRichards: Yes, I believe so.... assuming that this won't break existing sites... Also, this would validate any local security audits for TLS. TLS 1.0 was being phased out as early as 2018 and as of last year, no one really supports 1.0 or 1.1 anymore. I think we should drop any reference to 1.0 and imply that 1.2+ is all that should be referenced. Microsoft [stopped supporting](https://docs.microsoft.com/en-us/lifecycle/announcements/transport-layer-security-1x-disablement) both across all platforms from I can see.

misilot:
I tested with 

```yml
tls:
  options:
    default:
      minVersion: VersionTLS12
      cipherSuites:
        # TLS 1.0 - 1.2 cipher suites
        - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
        - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
        - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
        - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
        - TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
        - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305
```

It only offered TLS1.2 and TLS1.3, and received an "A".

benglis:
I have tested this, and was successful in removing TLS1 and TLS1.1 from use.  I was also able to get an "A" rating from ssllabs using this configuration.

DonRichards:
Instead of code comments can we have Tom's example added as code to test?

benglis:
I had tested this config on a running system, however, using this configuration with the make local-install-profile fails to curl the last few items (Homepage and Browse Collections pages) that are the last step of the setup...   

curl: (35) error:14094438:SSL routines:ssl3_read_bytes:tlsv1 alert internal error
make: *** [Makefile:385: local-install-profile] Error 35

DonRichards:
@g7morris Would you mind resolving the conflict?

g7morris:
@DonRichards I think I've fixed it as requested. Could you review again please?

g7morris:
I'm unclear at this point what of my original PR should be used? 

@DonRichards & @misilot What would you suggest?

marc1aj:
I add this code to traefik-tls.yml:

tls:
  options:
    default:
      minVersion: VersionTLS12
      cipherSuites:
        # TLS 1.0 - 1.2 cipher suites
        - TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256
        - TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
        - TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384
        - TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384
        - TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305
        - TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305

I then ran a make build and was successful in disabling TLS1 and TLS1.1. I also received an A rating from SSLLabs. I have not seen any adverse effects so far.

PR #265, opened 2022-06-15
Add custom Makefile

DonRichards:
This adds a check in the Makefile that looks for the existence of a custom.Makefile. If the file exist it will include it. Full instructions were added to the README. 

## To test
Using the instructions in the README you should be able to create a custom.Makefile and it should automatically recognize it when you run any make commands.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,11 @@ ENV_FILE=$(shell \
 	fi; \
 	echo .env)
 
+# If custom.makefile exists include it.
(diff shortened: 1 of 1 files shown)


ysuarez (approved):
This PR worked as expected. I approve that it gets merged.

ysuarez:
At today's tech call we thought of "changing the direction of composition," in other words have the custom Makefile include the project's default Makefile instead of the other way around like in this PR.

PR #274, opened 2022-07-13
Improve initial Prompt

DonRichards:
On a fresh install, there are several prompts. This is a bit confusing and doesn't lay out all of the options for generating secrets.

The new prompt looks like this
```shell
$ make up

Checking secrets...
  No secrets found in /home/user/isle-dc/secrets/live/

Everything (Drupal, SQL, Solr, etc.) needs passwords to run. These passwords are stored in the secrets directory.
  1. Generate new secrets
       ╰ This will generate random new passwords and output them into /home/user/isle-dc/secrets/live/
  2. Copy the existing secrets
       ╰ This will copy the existing secrets into the /home/user/isle-dc/secrets/live/. Every password will likely be 'password'.
  3. Create your own secrets
       ╰ This will copy the secrets from /home/user/isle-dc/secrets/template/ to /home/user/isle-dc/secrets/live/
            and will exit so you can edit the secrets manually prior to building the Docker image.
  4. Exit

Enter your choice: 3
Creating your own secrets. Copying templates to live directory.
```

## To test:
- Try each option. Starting in reverse order to speed up testing.
- Run `make clean` before each test so that the secrets/live/ directory is empty.
- Options # 3 and # 4 should exit the build process entirely when complete.
- Final test - without running `make clean` run `make down` then `make up`. It should completely bypass the above prompt.
- Options # 1 & # 2 should build completely and you should be able to log into Drupal with the password in secrets/live/DRUPAL_DEFAULT_ACCOUNT_PASSWORD

### DIFF
--- a/build/scripts/check-secrets.sh
+++ b/build/scripts/check-secrets.sh
@@ -68,12 +68,21 @@ function main() {
 	if [ ! "$(ls $(pwd)/secrets/live)" ]; then
 		echo -e "\n${YELLOW}Checking secrets...${RESET}"
 		echo "  No secrets found in $(pwd)/secrets/live/"
-		echo -e "\nThere are 2 basic methods to create secrets:"
-		echo " [1] - Generate new secrets via a script"
-		echo -e " [2] - Copy secrets from a $(pwd)/secrets/template directory into $(pwd)/secrets/live/ and then modify them\n"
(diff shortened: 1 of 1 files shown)


DonRichards:
Thanks @rosiel

PR #275, opened 2022-07-13
Add Github rate limit check

DonRichards:
When Github's rate limit is triggered it causes `fix-masonry` to silently fail with no obvious error messages. 

This adds the error message with the error occurs. The error message outputs the error and the time it should reset.

```shell
Github API rate limit exceeded. Skipping masonry fix.
  Either log into Github via the terminal or wait until Wed Jul 13 12:58:39 PM EDT 2022 for the rate limit to reset.
```


DonRichards:
Closing since there's no interest

PR #281, opened 2022-07-29
Rename the  islandora_install_profile_demo to demo

DonRichards:
The was supposed to be removed already. This fixes the issue with demo


noahwsmith:
Good catch, but really we want to use this variable rather than the hardcoded `demo` I think
https://github.com/Islandora-Devops/isle-dc/blob/development/sample.env#L132

DonRichards:
@noahwsmith Do you know where the logic is or determining was to execute for demo vs standard? Or is it demo vs local? Pointing to which docker-compose file to use, right?

noahwsmith:
`demo` and `local` are different, primarily in that `local` doesn't try to give you any test data, and `demo` has always come with some test data. Is that what you were asking?

PR #283, opened 2022-08-04
Fix make env 

DonRichards:
For some reason, this command is suddenly causing major permissions errors with the settings.php

Drupal resets to the install page instead of the home page.

## To test
Run `make env` and when it finishes you should be able to load the site.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -424,6 +424,12 @@ env:
 		$(MAKE) -B docker-compose.yml ; \
 		$(MAKE) pull ; \
 		$(MAKE) up ; \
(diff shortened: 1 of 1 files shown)


ajstanley:
Made a fresh box. changes .env ran make env
got a bunch of 
```
In Filesystem.php line 203:
                                                                          
  Failed to chmod file "/var/www/drupal/web/sites/default/settings.php".  
```
but then got` Fixing the error message:  In Filesystem.php line 203` and the box came back in perfect condition

PR #285, opened 2022-08-23
Update to use latest containers, fixes issue with houdini and ghostscript amongst others

noahwsmith:
While testing https://github.com/Islandora-Devops/islandora_install_profile_demo/pull/11 it became apparent that some of the minor fixes to the ISLE containers are required for the full stack to work, and there's no reason to stay on the ALPHA containers. This change would update all new projects to start with 1.0.7, not 1.0.0-alpha-16

### DIFF
--- a/sample.env
+++ b/sample.env
@@ -65,7 +65,7 @@ REPOSITORY=islandora
 
 # The version of the isle-buildkit images, non isle-buildkit images have
 # their versions specified explicitly in their respective docker-compose files.
-TAG=1.0.0-alpha-16
+TAG=1.0.7
 
 ###############################################################################
 # Exposed Containers & Ports


rosiel:
Great! Is there gonna be some kind of note in the documentation (here or elsewhere) to update this every time buildkit does a new tag?

noahwsmith:
For ISLE7 we do release notes by email, e.g. 
https://groups.google.com/g/islandora-isle/c/u-7zdyJtEKw

I defer to @g7morris and @nigelgbanks for how the release process will work going forward for ISLE 2.0.

g7morris:
Just reporting here that I just tested this now (today 8/30/2022) and it works great with no errors.

### Steps taken

* As notated in Noah's commit: change to the setting of `TAG=1.0.7` in the `sample.env`
* Run `make local`
* QCed local site with no errors found. 

cc @noahwsmith & @nigelgbanks @rosiel 

I don't know who needs to merge this in but I can't as I'm in the same org as @noahwsmith .

seth-shaw-asu:
@DonRichards will take a look; although folks on the Tech Call suggested that same-org folks can merge on Islandora Devops.

PR #286, opened 2022-09-09
This variable declaration has a typo in it

noahwsmith:
@nigelgbanks I think this should be a pretty open/shut case. Do you mind reviewing and merging?

### DIFF
--- a/build/docker-compose/docker-compose.traefik.yml
+++ b/build/docker-compose/docker-compose.traefik.yml
@@ -33,7 +33,7 @@ services:
       - --entryPoints.activemq.address=:8161
       - --entryPoints.solr.address=:8983
       - --entryPoints.code-server.address=:8443
-      - --log.level=${TRAEFIK_LOG_LEVEL-ERROR}
+      - --log.level=${TRAEFIK_LOG_LEVEL:-ERROR}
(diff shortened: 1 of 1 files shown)


build/docker-compose/docker-compose.traefik.yml
```diff
@@ -33,7 +33,7 @@ services:
       - --entryPoints.activemq.address=:8161
       - --entryPoints.solr.address=:8983
       - --entryPoints.code-server.address=:8443
-      - --log.level=${TRAEFIK_LOG_LEVEL-ERROR}
+      - --log.level=${TRAEFIK_LOG_LEVEL:-ERROR}
```
nigelgbanks, line 36: Seems both are acceptable and almost equivalent. 

> The extra : makes a difference only when parameter has been declared, but is null.

https://tldp.org/LDP/abs/html/parameter-substitution.html

So, the `:` will cover an additional case. It will now yield `ERROR` when `TRAEFIK_LOG_LEVEL` is defined and set to `null` not only when   `TRAEFIK_LOG_LEVEL` is not defined.

nigelgbanks (approved):

noahwsmith:
Thanks Nigel!

PR #291, opened 2022-09-20
Makefile: Abstract how we call sed, for compatibility with both Linux…

### LINKED ISSUE #290: make demo fails due to sed issue in "make demo_content" target
In a fresh install on ubuntu 20.04, "make demo" failed with
```
sed: can't read s/author_email\="mjordan@sfu"\,/author_email="mjordan@sfu", packages=["i7Import", "i8demo_BD", "input_data"],/g: No such file or directory
make[1]: *** [Makefile:374: demo_content] Error 2
make[1]: Leaving directory '/tmp/isle-dc'
make: *** [Makefile:334: demo] Error 2
```
this turned out to be a OSX vs. GNU sed incompatibility.

I've made a fix and will post a PR shortly.

jasonhildebrand:
… and OSX.  Also fix sed call so that it is idempotent.

This fixes #290 and increases readability/maintainability.

I have tested under Ubuntu 20.04, but I don't have access to OSX.  It should be tested there before merging.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,14 @@ ENV_FILE=$(shell \
 	fi; \
 	echo .env)
 
+# handle sed -i differences (see https://stackoverflow.com/a/4247319/3204023 for more context)
+ifeq ($(shell uname -s),Darwin)
+	SED_DASH_I=sed -i ''
+else  # GNU/Linux
+	SED_DASH_I=sed -i
+endif
+
+
 # If custom.makefile exists include it.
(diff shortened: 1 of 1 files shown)


DonRichards:
Testing this now, sorry for the delay

DonRichards:
No issues testing on Linux machine. Now testing on Mac

DonRichards (approved):
Tested and works! Thanks!

PR #315, opened 2022-11-29
Update tag to 1.0.10.

rosiel:
Updates the buildkit tag. (and that's it.)


rosiel:
done in https://github.com/Islandora-Devops/isle-dc/pull/316

PR #322, opened 2023-01-30
organized Makefile

joshdentremont:
I've been working with the Makefile a lot recently, and thought it would be nice to have all the different rules grouped by function. Previously things appeared to be in the order they had been added to the file, but since the makefile is now so large I thought this might be a useful change. I have grouped together the different rules for making a new site (demo, local, starter, starter_dev), put all the variables together, added groupings for backups, helpers, etc.

Nothing was removed, but everything has been moved around. I also added a few comments to rules that didn't have them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -202,7 +202,7 @@ help:
 	@echo ''
 	@echo 'Functions to build:'
 	# @grep '^.PHONY: .* #' Makefile | sed 's/\.PHONY: \(.*\) # \(.*\)/\1 \2/'
-	@awk '/^[a-zA-Z\-\_0-9]+:/ { \
+	@awk '/^[a-zA-Z\-_0-9]+:/ { \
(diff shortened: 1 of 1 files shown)


DonRichards:
@joshdentremont Are you able to resolve the merge conflicts?

DonRichards:
Screenshot 2023-03-15 at 11 53 13 AM
Looks like there's an error with `awk: cmd. line:1: warning: regexp escape sequence `\_' is not a known regexp operator` 

This may not be associated with your PR and is a leftover from a previous one but the issue is with the 2 lines with this. Removing this back slash worked on my local (mac)
```
@awk '/^[a-zA-Z\-\_0-9]+:/ { \

@awk '/^[a-zA-Z\-_0-9]+:/ { \
```
Would you mind adding it to this PR?

joshdentremont:
@DonRichards Done & done, thanks

PR #330, opened 2023-03-24
#328: Update demo_content to formulate the password for example_content.yml

ctgraham:
Aligns with 3rd party repo change to yaml syntax.
Escapes slashes in the user-supplied password.
Eliminates unneeded escaping for a colon.

Resolves #328.

PR #334, opened 2023-03-30
Add password and domain process improvements

DonRichards:
Continuing on efforts originally here https://github.com/Islandora-Devops/isle-dc/pull/330

The default Drupal password could contain characters that could be interpreted as syntax; this should help escape the characters.

This also adds the domain change to the demo content function.

## How to test
1. Copy secrets/template to secrets/live `cp -r secrets/template/* secrets/live/`
2. Add a semicolon "**;**" to the secrets/live/DRUPAL_DEFAULT_ACCOUNT_PASSWORD file
3. Copy the sample.env file to .env (`cp sample.env .env`) and change domain from **DOMAIN=islandora.traefik.me** to **DOMAIN=test.traefik.me**
4. Run `make starter_dev` or any other method to get isle-dc running
5. Open browser to https://test.traefik.me and verify the site loads.
6. Run `make demo_content`

It should complete without errors, the site should now show the imported content and if you check islandora_workbench/islandora_workbench_demo_content/example_content.yml file it should look like this
```yml
task: create
host: https://test.traefik.me/
username: admin
password: ";password"
input_dir: /workbench/islandora_workbench_demo_content/demo_content_files
input_csv: demo_content.csv
allow_missing_files: true
allow_adding_terms: true
media_use: "'Thumbnail Image'|'Original File'"
log_json: true
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -473,7 +473,7 @@ demo_content:
 	# if prod do this by default
 	[ -d "islandora_workbench" ] || (git clone https://github.com/mjordan/islandora_workbench)
 	cd islandora_workbench ; cd islandora_workbench_demo_content || git clone https://github.com/DonRichards/islandora_workbench_demo_content
-	$(SED_DASH_I) 's/^host.*/host: https:\/\/$(DOMAIN)\//g' islandora_workbench/islandora_workbench_demo_content/example_content.yml
+	$(SED_DASH_I) 's#^host.*#host: $(SITE)/#g' islandora_workbench/islandora_workbench_demo_content/example_content.yml
 	$(SED_DASH_I) 's/^password.*/password: "$(shell cat secrets/live/DRUPAL_DEFAULT_ACCOUNT_PASSWORD | sed s#/#\\\\\\\\/#g)"/g' islandora_workbench/islandora_workbench_demo_content/example_content.yml
 	cd islandora_workbench && docker build -t workbench-docker .
 	cd islandora_workbench && docker run -it --rm --network="host" -v $(QUOTED_CURDIR)/islandora_workbench:/workbench --name my-running-workbench workbench-docker bash -lc "./workbench --config /workbench/islandora_workbench_demo_content/example_content.yml"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -473,7 +473,8 @@ demo_content:
 	# if prod do this by default
 	[ -d "islandora_workbench" ] || (git clone https://github.com/mjordan/islandora_workbench)
 	cd islandora_workbench ; cd islandora_workbench_demo_content || git clone https://github.com/DonRichards/islandora_workbench_demo_content
-	$(SED_DASH_I) 's/^password.*/password: '$(shell cat secrets/live/DRUPAL_DEFAULT_ACCOUNT_PASSWORD | sed s#/#\\\\\\\\/#g)'/g' islandora_workbench/islandora_workbench_demo_content/example_content.yml
+	$(SED_DASH_I) 's/^host.*/host: https:\/\/$(DOMAIN)\//g' islandora_workbench/islandora_workbench_demo_content/example_content.yml
```
ctgraham: What if I also edit the `SITE` variable to use `http://` instead of `https://` in `.env`?
https://github.com/Islandora-Devops/isle-dc/blob/92ff14f71e20c5453ea308e61173061e8efc1031/sample.env#L130

Should this `sed` be based on `${SITE}` rather than on `${DOMAIN}` with an assumed protocol?

ctgraham (commented):
Should `${SITE}` be preferred in the config over `${DOMAIN}`?

Makefile
DonRichards: It was an issue with escaping the forward slashes. This was the simplest method to get it to work without having to modify the variable within the sed command.

ctgraham: You can alter the regexp delimiter as an easy solution:
```
$(SED_DASH_I) 's#^host.*#host: '${SITE}'#g' islandora_workbench/islandora_workbench_demo_content/example_content.yml
```
There should never be a `#` in a root URL.

DonRichards: @ctgraham This is one of those moments where I missed the simplest solution here. Thanks for the suggestion. That is a more elegant solution that will be much easier for others to understand. It's ready for review.

ctgraham (approved):

PR #336, opened 2023-04-14
Starter from codebase

joshdentremont:
Added ability to make starter with a different codebase.

In .env we can change the Packagist repo to something other than the starter site, or we can clone our own codebase folder and run `make starter` from that.

To test:
1. Run `make starter` as normal to get fresh starter site
2. `make clean`
3. Change the CODEBASE_PACKAGE variable in .env to another Packagist repo and run `make starter` to get a fresh install from that repo instead of the starter site
4. `make clean`
5. Put your own codebase folder in the isle-dc directory and run `make starter`. Instead of trying to pull from packagist it will just copy that into the container and run composer install

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -257,9 +257,13 @@ docker-compose.yml: $(SERVICES:%=build/docker-compose/docker-compose.%.yml) .env
 
 .PHONY: up
 .SILENT: up
-## Brings up the containers or builds demo if no containers were found.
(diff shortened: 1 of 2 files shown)


aOelschlager (approved):
I couldn't test the Packagist part because I don't have a codebase in there but I was able to test cloning a site from a github repository per your instructions and it ran great.

PR #337, opened 2023-05-18
added cache rebuild to database import

joshdentremont:
Running the database import command often messes up the site layout until the Drupal cache is cleared (looks like a CSS cache issue). This adds the drush command for clearing the cache to the database import process.

I'm not sure how to reliably trigger the need for the cache rebuild, because it doesn't seem to happen every time, but I have had it happen multiple times when importing the database from my production site to my local dev copy.


DonRichards (approved):
Worked, thanks

PR #340, opened 2023-05-30
add nginx variables for .env configuration

rbos:
We have a pretty big database and need to adjust the timeouts up from 60 seconds for some things. I had hardcoded these at one point, but they should really be variables.


joshdentremont:
@rbos We just discussed this on the tech call and it looks good. Could you also add the variables to sample.env?

Thanks

rbos:
Sure.

PR #341, opened 2023-05-31
update .env for nginx vars

rbos:
This should allow you to configure nginx timeouts via .env


joshdentremont:
Hi @rbos, we discussed this in the tech call today and it looks good. Would you mind adding the default values from buildkit to sample.env? 

So instead of the commented out options, it should look like this:

NGINX_FASTCGI_READ_TIMEOUT=60s
NGINX_FASTCGI_SEND_TIMEOUT=60s
NGINX_FASTCGI_CONNECT_TIMEOUT=60s
NGINX_PROXY_READ_TIMEOUT=60s
NGINX_PROXY_SEND_TIMEOUT=60s
NGINX_SEND_TIMEOUT=60s

PR #345, opened 2023-06-08
Update make build to specify platform

joshdentremont:
When building on an M1/M2 mac the default is to build for arm64 instead of amd64 which causes issues when trying to run a custom drupal image.

This PR specifies the image so that `make build` can be run on an M1/M2 mac. 

To test, run `make build` to create a custom drupal image, then test that the image doesn't throw errors when starting. Without this PR on an M1/M2 mac the drupal container will continually restart.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -262,7 +262,7 @@ build:
 	if [ ! -f $(PROJECT_DRUPAL_DOCKERFILE) ]; then \
 		cp "$(CURDIR)/sample.Dockerfile" $(PROJECT_DRUPAL_DOCKERFILE); \
 	fi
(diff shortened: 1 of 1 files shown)


DonRichards (approved):

joecorall:
Should this just be an entry in sample.env?

```
DOCKER_DEFAULT_PLATFORM=linux/amd64
```

joecorall (approved):
On an m2 I tried making the default linux/amd64 with

```
export DOCKER_DEFAULT_PLATFORM=linux/amd64
```

But that makes all the isle containers running amd64 which breaks things.

This override for just drupal LGTM

joshdentremont:
@joecorall we could put this in an environment variable, but I don't think the image works with anything other than linux/amd64. When I tried building with the default on my M2 it was giving me errors during startup, so I suspect that the Islandora Drupal image only supports amd

PR #350, opened 2023-07-12
Update make production

joshdentremont:
added commands to create drupal database and modify settings.php before running drush site install

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -198,6 +198,7 @@ production: generate-secrets
 	$(MAKE) pull
 	docker compose up -d --remove-orphans
 	docker compose exec -T drupal with-contenv bash -lc 'composer install; chown -R nginx:nginx .'
+	$(MAKE) drupal-database update-settings-php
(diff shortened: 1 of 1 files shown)


joshdentremont:
Before I added this line, make production was failing with:
```
docker compose exec -T drupal with-contenv bash -lc "drush si -y --existing-config minimal --account-pass 'XCIs6LUzks1MFwwU2WlFs1KDiOpz1gCsiDtNr8aAAqBQ9mjV'"

 You are about to:
 * Create a sites/default/settings.php file
 * CREATE the 'drupal' database.

 // Do you want to continue?: yes.

 [warning] Failed to drop or create the database. Do it yourself before installing. ERROR 2002 (HY000): Can't connect to server on '127.0.0.1' (115)


In SiteInstallCommands.php line 88:

  Existing config directory  not found
```

So I think it couldn't find the config/sync directory because it was not being specified in settings.php. 

For testing you should be able to run make production without this PR and it should throw a similar error. With this PR it should create the site successfully.

aOelschlager (approved):
I tested following the instructions for [installing a production server](https://islandora.github.io/documentation/installation/docker-custom/) and it worked.

PR #352, opened 2023-07-31
specify starter site version in sample.env

joshdentremont:
starter site needs to be installed at version 0.8.0 in order to be compatible with Isle 1.0.10.

Once we are ready to bump the default TAG to 2.0.0 or higher we will need to bump starter site as well to get Drupal 10.

### DIFF
--- a/sample.env
+++ b/sample.env
@@ -49,7 +49,7 @@ CUSTOM_IMAGE_TAG=latest
 
 # Packagist repo to use when creating a site with make starter
 # Change this if you want to build from a different codebase than the starter site
-CODEBASE_PACKAGE=islandora/islandora-starter-site:dev-main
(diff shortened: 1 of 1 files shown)


rosiel:
I think this should be `~0.8.0` which would allow any release up to (but not including) 1.0.0. This will take us out of the (IMO bad) habit of locking things down to specific versions which prevents updates from ever getting applied. 

(Generally we use the [caret operator](https://getcomposer.org/doc/articles/versions.md#caret-version-range-) e.g. ^0.8 but for pre-1.0.0 releases the caret treats every middle digit like its own major release)

rosiel:
On second thought, since this is breaking and there's unlikely to be another pre-1.0 release, I'll merge it with the vocal caveat that we should be specifying all-non-breaking-changes going forwards.

PR #353, opened 2023-08-17
Update docker compose in check-secrets.sh

joshdentremont:
the check secrets script still used docker-compose instead of docker compose.


aOelschlager (approved):
This works!

PR #359, opened 2023-10-03
Update traefik version

joshdentremont:
updated version of Traefik to match the version that site template uses.

### DIFF
--- a/build/docker-compose/docker-compose.traefik.yml
+++ b/build/docker-compose/docker-compose.traefik.yml
@@ -15,7 +15,7 @@ networks:
 services:
   traefik:
     restart: ${RESTART_POLICY:-unless-stopped}
-    image: traefik:2.7
+    image: traefik:2.8.3
     container_name: traefik
     # Do not set `api.insecure`, `api.dashboard`, `api.debug` to `true` in production.
     # Also do not expose database 3306/5432, as an entry point.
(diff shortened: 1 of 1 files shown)


joecorall (approved):
[LGTM](https://github.com/Islandora-Devops/isle-site-template/blob/afa3e82d7fec510c31f17c5788e024ce8c087cc9/docker-compose.yml#L595)

aOelschlager (approved):
LGTM too

g7morris:
Quick question of why aren't we attempting to use the latest version of traefik of [2.10.4](https://github.com/traefik/traefik/releases) ? Did someone test and that didn't work?

Or does this traefik [value](https://github.com/Islandora-Devops/isle-site-template/blob/afa3e82d7fec510c31f17c5788e024ce8c087cc9/docker-compose.yml#L595) in the isle-site-template also need to be upgraded first and then this project? 

Just curious what the order of ops is here?

Should we work with @nigelgbanks on this too?

g7morris:
Just an FYI, tested running with latest `2.10.4` on an existing site with no ill effects. More extensive testing warranted.

joshdentremont:
@g7morris That's a good question. I just used the same one as the site template because I figured that was what was tested with Buildkit. 

I would be happy to bump it up, but I think we should probably keep Isle and Site Template on the same version, so we should bump both at the same time.

nigelgbanks:
No need to sync with `isle-site-template`. The version used in the site template was just the latest available at the time it was written, I think. There isn't any real need to test with `isle-buildkit` releases, either. The only concern with `traefik` is the Docker labels used in the docker-compose.yml files. Since `isle-site-template` and `isle-dc` have different docker-compose.yml files, there isn't any need to keep them in sync.

joshdentremont:
Updated this to use the newest version, 2.10.5

PR #366, opened 2023-10-26
Make the chown nginx to tolerate when a file is set to archive/read.

DonRichards:
Presently, within the codebase, there's a minor permissions hiccup involving select files in the .git/ folder. While adjusting their permissions isn't crucial for building or running the site, it's currently causing an error when specific lines in the makefile attempt to alter ownership. This update tackles that issue, enabling them to gracefully handle these permissions hiccups without interrupting the flow. 🚀🛠️

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,7 +158,7 @@ local: generate-secrets
 		echo "Waiting for /var/www/drupal directory to be available..."; \
 		sleep 2; \
 	done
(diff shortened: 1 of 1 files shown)


aOelschlager (changes requested):
This works. I was able to install without issue.
Just needs to pull in the newest development commits to get back in sync.

aOelschlager (approved):

PR #367, opened 2023-11-15
Add php 8.2 // isle-buildkit:3.x support

joecorall:
## Summary

https://github.com/Islandora-Devops/isle-buildkit/pull/299 bumped the isle-buildkit major version from `2.x` to `3.x` since that PR upgraded PHP `8.1 -> 8.2`, and 8.2 could break drupal 9 sites. This PR simply checks if the tag is 3.x and sets the PHP FPM pid file to the correct value. It also bumps the starter site and default isle-buildkit tags to the latest versions.

### New functionality

Running this command (which will destroy any site you may have running locally) spins up a d10 site with php 8.2

```
make down clean
make starter TAG=3.0.0
```

### Regression test

Running this command (which will destroy any site you may have running locally) spins up a d10 site with php 8.1

```
make down clean
make starter TAG=2.0.10
```


aOelschlager (approved):
Tested with 3.0.0 tag and 2.0.10 tag and both ran without error. This works, thanks Joe!

PR #376, opened 2024-02-12
374-update-images Update to BuildKit images 3.1.

alxp:
This was suggested in the last tech call for getting JP2 support to work out of the box.

I've tested it and I can now upload JP2s and they appear in Mirador.

image

To test simply build with 'make starter' or 'make starter_dev', then create a new Repository Item object, and give it the model "Paged Content."

Then go to the Children tab, and choose 'Bulk add children".

On the next page, choose Repository Item for the content type, 'Page' for the model, 'File' for the media type and 'Original File' for the media use tag.

Then on the neext page, upload one or more JP2 files and click 'Finish.

When you go back to the node's View tab, the pages should appear in the Mirador viewer.

### DIFF
--- a/sample.env
+++ b/sample.env
@@ -94,7 +94,7 @@ REPOSITORY=islandora
 
 # The version of the isle-buildkit images, non isle-buildkit images have
 # their versions specified explicitly in their respective docker-compose files.
-TAG=3.0.0
+TAG=3.1.0
 
 ###############################################################################
 # Exposed Containers & Ports
--- a/sample.env
+++ b/sample.env
@@ -1,6 +1,6 @@
 # Environment variables defined in this file apply to both the Makefile and to
 # docker-compose.yml
-# 
+#
 # Due to restrictions in the `env-file` format we cannot specify multi-line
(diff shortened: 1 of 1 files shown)


sample.env
```diff
@@ -94,7 +94,7 @@ REPOSITORY=islandora
 
 # The version of the isle-buildkit images, non isle-buildkit images have
 # their versions specified explicitly in their respective docker-compose files.
-TAG=3.0.0
+TAG=3.1.0
```
joecorall:
```suggestion
TAG=3.1.2
```

[3.1.0](https://github.com/Islandora-Devops/isle-buildkit/releases/tag/3.1.0) upgrades solr server from 8.x to 9.x... If we're bumping the isle-buildkit tag for Cantaloupe support, I suggest we pin to [3.1.2](https://github.com/Islandora-Devops/isle-buildkit/releases/tag/3.1.2) which is the first point release for cantaloupe in several years and resolves several security issues.

joecorall (approved):

Issue #379, opened 2024-03-07
Network gateway not found errors


aOelschlager:
I have not experienced this error myself, however, four people have now ran into this error and reported it on Slack. I'm not sure why it's happening but the docker compose file that is generated is declaring the gateway network as external and that causes the network to not be created thus giving this error. 
I know two of these were encountered on a vm. One had docker compose version one still installed. 

Temporary fix has been to edit the docker compose file and remove the external part for the gateway network. Next, comment out the line that generates the docker compose file in the Makefile. Then run make starter or starter_dev again.

alxp:
I don't know enough about docker to speak confidently about what the correct thing to chagne it is, but another temporary workaround is to run the command 

```bash
docker network create gateway
```

and delete it with

```bash
docker network rm gateway
```

when not needed.

ajstanley:
The problem seems to come from declaring external: true in the network definition instead of internal: false
[Docker docs.](https://docs.docker.com/reference/cli/docker/network/create/)

joecorall:
Resolved by https://github.com/Islandora-Devops/isle-dc/pull/381

PR #381, opened 2024-03-13
Updated network option

ajstanley:
This relates to [issue 379](https://github.com/Islandora-Devops/isle-dc/issues/379).
External is not a recognized Docker network option, but it only fails on some machines.  This ought to work anywhere, but its only been tested on a Mac, and an old one at that,

### DIFF
--- a/build/docker-compose/docker-compose.activemq.yml
+++ b/build/docker-compose/docker-compose.activemq.yml
@@ -3,7 +3,7 @@ networks:
   default:
     internal: true
   gateway:
-    external: true
+    internal: false
 volumes:
   activemq-data:
 services:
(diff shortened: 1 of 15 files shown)


misilot:
That's weird since external is defined in the documentation: https://docs.docker.com/compose/compose-file/06-networks/#external

misilot:
Though it looks like the default is `internal: false` so would it be worth removing that config option all together?

aOelschlager:
I would agree with @misilot that internal:false can be removed from the yaml files. In the tech call it was mentioned that the original purpose for this (external:true) was to be able to run isle 7 at the same time as isle-dc. I don't know if anyone is using it this way anymore but I highly doubt it.

joecorall (approved):
Before this PR I was getting

```
$ make starter
...
...
docker compose up -d --remove-orphans
Error response from daemon: network gateway not found
make: *** [starter] Error 1

$ echo $?
2
```


After this PR `make starter` gets past the `docker compose up -d --remove-orphans` command successfully.

PR #383, opened 2024-04-17
Set default isle-buildkit tag to 3.2.1 for php83

### LINKED ISSUE #382: Update tag and add php-fpm83 to makefile
Buildkit has been updated to use php 8.3 for tags 3.2 and up. [Sample.env ](https://github.com/Islandora-Devops/isle-dc/blob/development/sample.env#L97)and the [Makefile](https://github.com/Islandora-Devops/isle-dc/blob/development/Makefile#L117-L122) need to be updated.

joecorall:
Resolves #382


aOelschlager (approved):
Looks good. Thank you Joe!

PR #385, opened 2024-04-19
Add Windows test support

joecorall:
Lets see if we can add Windows test support here.

## Note on MacOS tests

Until [M1 runners have docker support](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md) getting mac os tests to run on modern hardware is not possible. And adding docker support to a `macos-13` runner and getting isle-dc running takes over 30min so not feasible to test.

But lets

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,12 +21,12 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


joecorall:
Closing this for now since we do not have isle-buildkit windows support

PR #386, opened 2024-04-19
Fixup for waiting for islandora.traefik.me loop

pennedav:
The wait-for-drupal-locally Makefile target gets stuck in an infinite loop if you have modified the DOMAIN environment setting from the default islandora.traefik.me.

This pull request fixes that issue and uses the DOMAIN setting from the user .env file.


joecorall (approved):
Thanks for the PR! Tested locally and this is not a regression for `islandora.traefik.me` so trusting this works as advertised for production domains.

PR #390, opened 2024-04-24
Add test for config export through UI

joecorall:
Accessing a config export through the Drupal admin UI fails with the default timeout settings in this repo (i.e. hitting this URL >> https://islandora.traefik.me/admin/config/development/configuration/full/export-download << after `make starter` results in a 5XX error)

Added a test around this proving a set of timeout environment variables are needed to fix the issue.

Also made the nightly build check against the latest version of `isle-buildkit`

## Before

https://github.com/Islandora-Devops/isle-dc/actions/runs/8824402956/job/24226804416?pr=390

## After

https://github.com/Islandora-Devops/isle-dc/actions/runs/8824446972/job/24226928240

## TODO

Figure out exactly which environment variables we need and only set those.

## Links

https://islandora.slack.com/archives/CM6F4C4VA/p1713892169660929?thread_ts=1713892169.660929&cid=CM6F4C4VA


ajstanley (approved):
These changes worked for me in two different environments.
In both cases I could download config, but in both cases files were left behind in the /tmp directory because Drupal was under the mistaken impression that the files weren't there.
That appears to be a separate issue.

joecorall:
> In both cases I could download config, but in both cases files were left behind in the /tmp directory because Drupal was under the mistaken impression that the files weren't there.

The file being left behind is intended behavior. [There is no cleanup that happens after the export](https://git.drupalcode.org/project/drupal/-/blob/dae385398eb9860650b50f2aac087c33ee35083b/core/modules/config/src/Controller/ConfigController.php#L155)

[There is an attempt to remove the file before the export happens](https://git.drupalcode.org/project/drupal/-/blob/dae385398eb9860650b50f2aac087c33ee35083b/core/modules/config/src/Controller/ConfigController.php#L135). This is the cause of the mysterious watchdog entries. When you run the export for the first time there will not be a file there, so that cleanup call will log in watchdog. Then the export runs with no cleanup happening after the fact resulting in puzzled engineers.

Going to issue a PR on Drupal core for this if one doesn't already exist. We've spent several hours tracking this down thinking we may have some odd `/tmp` side effect in our stack and I know from the tech call @DonRichards has ran into this quandry, too. And a google search says many others have, too.

🤞 https://www.drupal.org/project/drupal/issues/3443874

PR #396, opened 2024-06-03
Remove matomo

joecorall:


alxp (approved):

alxp:
Building of the newest buildkit tag is broken without this and I just tested this PR and everything builds as expected.

Issue #401, opened 2024-08-05
Update-settings-php doesn't work with Drush 13


aOelschlager:
I was testing the Drush 13 pr for the starter-site and the update-settings.php make function is no longer working. When I removed that call and just use the same default_settings.txt and environment variables as the isle-site-template it worked. Perhaps this would be a good time to change this function to be more inline with how the isle-site-template implements the settings.php file.

PR #403, opened 2024-08-07
Update how settings.php gets appended

aOelschlager:
For issue https://github.com/Islandora-Devops/isle-dc/issues/401

This removes the update-settings make command from the starter variants and replaces the default_settings.txt in the starter site. This is how the isle site template does it's configuration.

To test run make starter_dev or make starter.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -558,7 +558,7 @@ delete-shortcut-entities:
 .PHONY: hydrate
 .SILENT: hydrate
 ## Reconstitute the site from environment variables.
-hydrate: update-settings-php update-config-from-environment solr-cores namespaces run-islandora-migrations
+hydrate: update-config-from-environment solr-cores namespaces run-islandora-migrations
 	docker compose exec -T drupal drush cr -y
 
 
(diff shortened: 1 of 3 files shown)


aOelschlager:
Make starter_dev works but make starter errors out.

alxp:
Confirm that make starter_dev works for me.

make starter seems to fail on Drupal install looking for the database, but I was able to immediately docker compose exec mariadb with-contenv bash and run mysql and see the database was created. Running make starter again continues the build. So I don't think the error is related to the problem addressed in this PR>

aOelschlager:
That fixed the error on make starter. This works but it's still calling the update-settings make command during make hydrate. If I remove that command from make hydrate then make starter and make starter_dev white screen on adding repository objects or media. I think it has to do with the fedora url not being set. I should have that fixed here soon.

aOelschlager:
This is ready for review now.

alxp (approved):
Just tested make starter and make starter_dev and they both worked.

PR #404, opened 2024-08-09
Remove docker compose version in the yaml files.

aOelschlager:
Including the version number in the docker compose yaml files throws a docker compose version is obsolete message when running make starter/starter_dev. Docker compose v2 will always use the latest version of the compose file scheme and it no longer requires the version number.

To test, run make starter or starter_dev and observe that there is no longer the version obsolete message.


alxp (approved):
Does what it says on the tin.

PR #410, opened 2024-10-24
Set DRUSH_OPTIONS_URI in Drupal container

xurizaemon:
Without URI set, Drush defaults to `http://default`.

```
69e89db055bd:/var/www/drupal# drush status --fields='Site URI'
Site URI : http://default 
```

Since we set `DRUSH_DEFAULT_SITE_URL`, we can populate `DRUSH_OPTIONS_URI` too and save some unnecessary `--uri`.

```
69e89db055bd:/var/www/drupal# export DRUSH_OPTIONS_URI=${DRUPAL_DEFAULT_SITE_URL} 
69e89db055bd:/var/www/drupal# drush status --fields='Site URI'
Site URI : http://islandora.example.org
# and for drush user:login, etc.
```


alxp (approved):
Does what it says.

PR #425, opened 2025-06-27
Update Makefile

kwisatz:
Corrects two locations where `islandora.dev` is hardcoded and would not respect the DOMAIN variable defined in the .env, making the `starter` process fail.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -621,8 +621,8 @@ starter-finalize:
 
 	# see https://github.com/Islandora-Devops/isle-site-template/pull/76
 	@CACHE_KEY=$$(docker compose exec -T drupal with-contenv bash -lc "/var/www/drupal/web/core/scripts/rebuild_token_calculator.sh 2>/dev/null") ; \
-	echo "curl https://islandora.dev/core/rebuild.php?$$CACHE_KEY"; \
-	docker compose exec -T drupal with-contenv bash -lc "curl -sLo /dev/null \"https://islandora.dev/core/rebuild.php?$$CACHE_KEY\""
+	echo "curl https://$(DOMAIN)/core/rebuild.php?$$CACHE_KEY"; \
+	docker compose exec -T drupal with-contenv bash -lc "curl -sLo /dev/null \"https://$(DOMAIN)/core/rebuild.php?$$CACHE_KEY\""
 
 	docker compose exec -T drupal with-contenv bash -lc "drush cr"
 	docker compose exec -T drupal with-contenv bash -lc "drush -l $(SITE) user:role:add fedoraadmin admin"


Makefile
```diff
@@ -731,7 +731,7 @@ compose-up:
 .PHONY: wait-for-drupal-locally
 .SILENT: wait-for-drupal-locally
 wait-for-drupal-locally:
-	while ! curl -s -o /dev/null -m 5 https://$(DOMAIN)/ ; do \
+	while ! curl -ks -o /dev/null -m 5 https://$(DOMAIN)/ ; do \
```
joecorall: `-k` would imply the TLS cert is not trusted by the host OS and might be a signal of an incomplete install?

kwisatz: Yes, maybe let me play a bit longer with this or if you could answer the following questions, that could help:
I can see that at the beginning of the process, self-signed certs are generated for `islandora.dev`, etc… I assume the CA used here is part of the CA available to curl or the container from which curl is run for most of these tasks.

```
The local CA is already installed in the system trust store! 👍


Created a new certificate valid for the following names 📜
 - "*.islandora.dev"
 - "islandora.dev"
 - "*.islandora.io"
 - "islandora.io"
 - "*.islandora.info"
 - "islandora.info"
 - "localhost"
 - "127.0.0.1"
 - "::1"
```

Is the `starter` task not supposed to be used with a domain other than islandora.dev?
~~Is `traefik`'s ACME only invoked for the `production` task?~~

I can also see other problems that prevent the certificate from being generated, such as

```
traefik  | 2025-06-27T17:20:18Z ERR Unable to obtain ACME certificate for domains error="cannot get ACME client acme: error signing eab content: failed to External Account Binding sign content: go-jose/go-jose: invalid key size for algorithm" ACME CA=https://acme-v02.api.letsencrypt.org/directory acmeCA=https://acme-v02.api.letsencrypt.org/directory domains=["islandora.my.tld"] providerName=myresolver.acme routerName=minisle-cantaloupe_https@docker rule="Host(`islandora.my.tld`) && PathPrefix(`/cantaloupe`)"
```

I will need to check where in the process traefik starts the ACME process and where this invalid key size is coming from.
All in all, this PR may have been a bit premature, feel free to dismiss and we'll create another one once we have a firmer grasp of the `isle-dc` setup.

kwisatz: OK, I figured out the source of the EAB signature error.
The generated docker-compose file contains parameters being passed to traefik with eab.kid and eab.hmac set to empty values and it does not like that. 
I suppose either these arguments should not be added if empty in the `.env` or the comments in the `.env` should specific that these are not optional if `USE_ACME`  is set to a truthy value.

I'll now run this again without the `-k` flag to see if the cert will be generated early enough for the curl requests to be run against a valid cert.

kwisatz:
I've removed the `-k` flags to curl. I think for the eab issue it'd be best to discuss this in a ticket first.

joecorall (approved):

